PR #54, opened 2012-12-18
Generic Instances for HashMap and HashSet

2piix:
Used GHC's "DeriveGeneric" extension to derive Generic instances for HashSets and HashMaps.

### DIFF
--- a/Data/HashMap/Base.hs
+++ b/Data/HashMap/Base.hs
@@ -1,5 +1,9 @@
 {-# LANGUAGE BangPatterns, CPP, DeriveDataTypeable, MagicHash #-}
-{-# OPTIONS_GHC -fno-full-laziness -funbox-strict-fields -XDeriveGeneric #-}
+{-# OPTIONS_GHC -fno-full-laziness -funbox-strict-fields #-}
+
+#if defined(__GLASGOW_HASKELL__)
+{-# LANGUAGE DeriveGeneric #-}
+#endif
 
 module Data.HashMap.Base
     (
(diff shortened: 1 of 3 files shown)


unordered-containers.cabal
```diff
@@ -39,6 +39,7 @@ library
   build-depends:
     base >= 4 && < 5,
     deepseq >= 1.1,
+    ghc-prim >= 0.2,
```
jaspervdj: @2piix, why is this dependency necessary?

jaspervdj:
+1

Data/HashMap/Base.hs
```diff
@@ -1,6 +1,10 @@
 {-# LANGUAGE BangPatterns, CPP, DeriveDataTypeable, MagicHash #-}
 {-# OPTIONS_GHC -fno-full-laziness -funbox-strict-fields #-}
 
+#if defined(__GLASGOW_HASKELL__)
```
tibbe: You can remove this ifdef. We only need to support the last 3 major releases (so that's back to GHC 7.4) and this pragma existed since 7.2. There are no other compilers than GHC. :)

unordered-containers.cabal
tibbe: `GHC.Generics` is exported by base so it shouldn't be necessary.

phadej: Not for all `base`, `GHC.Generics` is in only since `base-4.6`. Also `Generics` are added in `GHC-7.2`, yet `unordered-containers` seems to compile down to `7.0`.

I'm not sure whether @tibbe wants to keep the compatibility.

treeowl:
These types are and have always been exported abstractly. We really don't want `Generic` instances!

Issue #59, opened 2013-02-19
`mapMaybe` and `mapMaybeWithKey`?


favonia:
I'm wondering if we can have `mapMaybe` and `mapMaybeWithKey`. (c.f. [`IntMap`](http://hackage.haskell.org/packages/archive/containers/0.5.2.1/doc/html/Data-IntMap-Strict.html#v:mapMaybe)) This is equivalent to

```fromList . mapMaybe (uncurry f) . toList```

but I guess it's slower. It seems that `filter` does something quite efficient and it'd be nice to have other functions enjoying the same efficiency.

PR #94, opened 2014-12-14
Add mapMaybe and mapMaybeWithKey (issue #59)

jcpetruzza:
They share the implementation with filter and filterWithKey, without affecting their structure reuse.

This fixes issue #59.


tibbe:
Merged, thanks!

PR #95, opened 2015-01-09
Implement Foldable.foldl' directly.

S11001001:


S11001001:
Withdrawn; not compatible with GHC 7.4.2.

PR #97, opened 2015-01-19
Derive Generic instances

jaspervdj:
I rewrote #54 so it applies cleanly to the current master, and gets the GHC version requirements correct.


jaspervdj:
Never mind, this is not relevant as long as we don't have an internal module exporting the constructors anyway. Sorry for the spam.

Issue #78, opened 2014-04-01
Suggestion: provide Hashable instances for containers themselves


nikita-volkov:

phadej:
Yes, I found https://issues.apache.org/jira/browse/THRIFT-1927 and that implementation is just so incorrect, it makes me very sad :(

phadej:
Also: don't open issues on April's Fool day.

andrewthad:
I think this can be closed.

treeowl:
Closed in #111

PR #111, opened 2015-09-24
Hashable instances fix

phadej:
Should be ready to review.

### DIFF
--- a/Data/HashMap/Base.hs
+++ b/Data/HashMap/Base.hs
@@ -98,6 +98,7 @@ import Control.DeepSeq (NFData(rnf))
 import Control.Monad.ST (ST)
 import Data.Bits ((.&.), (.|.), complement)
 import Data.Data hiding (Typeable)
+import Data.Ord (comparing)
 import qualified Data.Foldable as Foldable
 import qualified Data.List as L
 import GHC.Exts ((==#), build, reallyUnsafePtrEquality#)
@@ -201,9 +202,6 @@ instance Traversable (HashMap k) where
 instance (Eq k, Eq v) => Eq (HashMap k v) where
     (==) = equal
 
-instance (Hashable k, Hashable v) => Hashable (HashMap k v) where
-    hashWithSalt = foldlWithKey' (\h k v -> h `H.hashWithSalt` k `H.hashWithSalt` v)
-
 equal :: (Eq k, Eq v) => HashMap k v -> HashMap k v -> Bool
 equal t1 t2 = go (toList' t1 []) (toList' t2 [])
   where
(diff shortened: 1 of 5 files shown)


phadej:
Resolves https://github.com/tibbe/unordered-containers/issues/78

tests/HashMapProperties.hs
```diff
@@ -49,6 +49,14 @@ pFoldable :: [(Int, Int)] -> Bool
 pFoldable = (L.sort . Foldable.foldr (:) []) `eq`
             (L.sort . Foldable.foldr (:) [])
 
+pHashable :: [(Key, Int)] -> [Int] -> Int -> Property
+pHashable xs is salt =
+    x == y ==> hashWithSalt salt x === hashWithSalt salt y
+  where
+     ys = L.map snd . L.sort . L.zip (is ++ [L.maximum (0:is) + 1..]) $ xs
```
tibbe: Would it be better to randomly shuffle `ys`?

tests/HashMapProperties.hs
```diff
@@ -49,6 +49,14 @@ pFoldable :: [(Int, Int)] -> Bool
 pFoldable = (L.sort . Foldable.foldr (:) []) `eq`
             (L.sort . Foldable.foldr (:) [])
 
+pHashable :: [(Key, Int)] -> [Int] -> Int -> Property
+pHashable xs is salt =
+    x == y ==> hashWithSalt salt x === hashWithSalt salt y
```
tibbe: Surprised this holds if `xs` ever has duplicate keys, as e.g. `fromList [(1,2),(1,3)]` and `fromList [(1,3),(1,2)]` don't result in the same map.

phadej: But it does, this is the point of this PR!

tests/HashMapProperties.hs
phadej: I actually to randomly shuffle `xs` into `ys`, I'll give that shuffle stuff a name and a comment

tests/HashMapProperties.hs
tibbe: So this hashable instance creates collisions for different maps on purpose?

phadej: Well, *different*. `fromList [(1,2),(1,3)]` and `fromList [(1,3),(1,2)]` are the same if you ask from `(==)`. And there is a `Hashable` law:

> If two values are equal according to the `==` method, then applying the `hashWithSalt` method on each of the two values must produce the same integer result if the same salt is used in each case.

phadej: The hashes are still (try to be) different for genuinely different maps:
```hs
λ *Data.HashMap.Lazy > hash (fromList [(1,2),(3 :: Int,4 :: Int)])
4712185439596716192
λ *Data.HashMap.Lazy > hash (fromList [(1,2),(3 :: Int,5 :: Int)])
4712185439596716193
```

phadej: Ah, no I understand. The same key.

phadej: That why there is an implication `x == y ==> `

phadej:
Clarified the test. I don't know better way to have a list and randomly shuffled version of it for the property.

phadej:
@tibbe Is there something still needing a fix or an improvement?

Data/HashMap/Base.hs
```diff
@@ -221,11 +218,32 @@ equal t1 t2 = go (toList' t1 []) (toList' t2 [])
...
-    toList' l@(Leaf _ _)          a = l : a
-    toList' c@(Collision _ _)     a = c : a
-    toList' Empty                 a = a
+instance (Hashable k, Hashable v) => Hashable (HashMap k v) where
+    hashWithSalt salt hm = go salt (toList' hm [])
+      where
+        go :: Int -> [HashMap k v] -> Int
+        go s [] = s
+        go s (Leaf _ l : tl)      = s `hashLeafWithSalt` l `go` tl
+        -- For collisions we hashmix hash value, and then array of values' hashes sorted
```
tibbe: Please keep line length <= 80 chars. Here and elsewhere.

tests/HashMapProperties.hs
tibbe: Lets see if I can paraphrase the test to see if I understand what it's doing. The test is

 * checking that two equal hash maps hash to the same value, where
 * given that they were created from the same list (which was permuted in one case).

Correct?

Given the way we generate the input list, will we end up throwing away most of the generated lists (by using `==>`) because they contain duplicate keys but different values? I guess that's fine as long as QC doesn't give up generating input (which it does if it can't find 100 cases after generating 500, I believe.)

phadej: Correct.

QuickCheck doesn't gave up for me, yet I could change this to
```hs
pHashable zs is salt =
    x == y ==> hashWithSalt salt x === hashWithSalt salt y
  where
    xs = Map.fromList . Map.toList $ zs -- Drop duplicate keys
    ys = shuffle is xs
```

Then `x == y ==>` won't be strictly required, but i'd keep as it communicates the intention.

tibbe: Lets keep it as is. Fix the line length and I'll merge.

phadej:
All lines are under 80 characters now.

PR #115, opened 2015-12-31
Export HashSet constructor and accessor

Philonous:
It would be useful to use HashSets as key sets (for difference / intersection oprations). Unfortunately there doesn't seem to be a cheap way to do this now, but it seems that exposing the newtype constructor would achieve this.


Philonous:
This is actually already implemented by 367b358f753f36dfce05f96793189de904863542

PR #132, opened 2016-05-09
Minor haddock elaboration

### LINKED ISSUE #131: Order of toList
Am I correct assuming that `Data.HashMap.Strict.toList` returns hashmap key-value pairs always in sorted order as a result of hashmap internal structure? If so probably it's worth a note in haddocks about it?

qrilka:
Fixes #131

### DIFF
--- a/Data/HashMap/Base.hs
+++ b/Data/HashMap/Base.hs
@@ -1051,7 +1051,7 @@ elems = L.map snd . toList
 -- ** Lists
 
 -- | /O(n)/ Return a list of this map's elements.  The list is
--- produced lazily.
+-- produced lazily. The order of its elements is unspecified.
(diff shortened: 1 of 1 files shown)


Data/HashMap/Base.hs
```diff
@@ -1051,7 +1051,7 @@ elems = L.map snd . toList
 -- ** Lists
 
 -- | /O(n)/ Return a list of this map's elements.  The list is
--- produced lazily.
+-- produced lazily. The order of its elements is unspecified
```
tibbe: Nitpick: end sentences in a dot.

qrilka: oops, thanks, fixed

PR #136, opened 2016-06-20
Don't use -fregs-graph

bgamari:
This option has been somewhat broken and therefore 
disabled for the last several GHC releases. See GHC #7679.


bgamari:
@tibbe, could you push a release with this patch?

PR #140, opened 2017-01-02
Add instance for Data.Functor.Classes

phadej:


phadej:
`Hashable1` and `Hashable2` tested using my branch for `hashable`.

CHANGES.md
```diff
@@ -1,3 +1,11 @@
+## 0.2.8.0
+
+ * Add `Eq1/2`, `Show1/2`, `Read1` instances with `base-4.9
```
tibbe: Missing trailing `

CHANGES.md
```diff
@@ -1,3 +1,11 @@
...
+
+ * `Eq (HashSet a)` doesn't require `Hashable a` anymore, only `Eq a`.
```
tibbe: This is a bit risky because if we ever need the `Hashable` constraint in the future in a new implementation of `Eq`, adding it is a breaking change. Perhaps just leave as is?

phadej: It was motivated by `Eq1` instance, as you cannot use `Hashable`, but I can leave the `Eq` instance as it is.

phadej:
ping @tibbe

Issue #144, opened 2017-02-07
Time complexity of `HashSet.member` is O(log(n))?


vaibhavsagar:
The time complexity of `member` is stated to be _O(min(n,W))_ but its implementation is identical to `HashMap.member` whose time complexity is stated as _O(log n)_. Is the documentation incorrect or is there something I'm missing?

tibbe:
That's a bug. It should be O(log n). This is a left-over from the previous implementation.

PR #145, opened 2017-02-08
Update documented time complexity.

vaibhavsagar:
According to #144, this documentation was a holdover from the previous implementation.


vaibhavsagar:
Thanks!

Issue #137, opened 2016-08-17
Add differenceWith?


joe9:
Hello,

Thanks for sharing the library.

Just want to check if you could add differenceWith as provided by Data.Map.Strict.

Thanks again

mschristiansen:
Can be closed since fixed in #146 .

PR #146, opened 2017-03-11
Add `differenceWith` function.

mschristiansen:
Same function as in Data.Map and with tests against that function. Let me know if there is anything I've missed or need to fix.

### DIFF
--- a/Data/HashMap/Base.hs
+++ b/Data/HashMap/Base.hs
@@ -45,6 +45,7 @@ module Data.HashMap.Base
 
       -- * Difference and intersection
     , difference
+    , differenceWith
     , intersection
     , intersectionWith
     , intersectionWithKey
@@ -933,6 +934,18 @@ difference a b = foldlWithKey' go empty a
                  Nothing -> insert k v m
                  _       -> m
 {-# INLINABLE difference #-}
+
+-- | /O(n*log m)/ Difference with a combining function. When two equal keys are
+-- encountered, the combining function is applied to the values of these keys.
+-- If it returns 'Nothing', the element is discarded (proper set difference). If
+-- it returns (@'Just' y@), the element is updated with a new value @y@.
+differenceWith :: (Eq k, Hashable k) => (v -> w -> Maybe v) -> HashMap k v -> HashMap k w -> HashMap k v
+differenceWith f a b = foldlWithKey' go empty a
+  where
+    go m k v = case lookup k b of
+                 Nothing -> insert k v m
+                 Just w  -> maybe m (\y -> insert k y m) (f v w)
+{-# INLINABLE differenceWith #-}
 
 -- | /O(n*log m)/ Intersection of two maps. Return elements of the first
 -- map for keys existing in the second.
--- a/Data/HashMap/Lazy.hs
+++ b/Data/HashMap/Lazy.hs
@@ -64,6 +64,7 @@ module Data.HashMap.Lazy
 
       -- * Difference and intersection
     , difference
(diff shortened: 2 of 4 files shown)


mschristiansen:
Fixes #137

Data/HashMap/Base.hs
```diff
@@ -934,6 +935,18 @@ difference a b = foldlWithKey' go empty a
...
+-- | /O(n*log m)/ Difference with a combining function. When two equal keys are
+-- encountered, the combining function is applied to the values of these keys.
+-- If it returns 'Nothing', the element is discarded (proper set difference). If
+-- it returns (@'Just' y@), the element is updated with a new value @y@.
+differenceWith :: (Eq k, Hashable k) => (v -> w -> Maybe v) -> HashMap k v -> HashMap k w -> HashMap k v
+differenceWith f a b = foldlWithKey' go empty a
+  where
+    go m k v = case lookup k b of
+                 Nothing -> insert k v m
+                 Just w  -> maybe m (\y -> insert k y m) (f v w)
```
tibbe: This bit is tricky. In the strict case you need to use `Data.HashMap.Strict.insert` to make sure the value is evaluated before being inserted in the map. You will need to duplicate this function into that module and use the right `insert`. If you look in that module you will see that happens for other functions (including `insert`) that need to force the value.

mschristiansen: I have added the function to `Data.HashMap.Strict`. There is a number of functions in `Strict` that are just reexported from `Base` although they also make use of `insert` and similar. Should `difference` as an example actually be redefined in `Strict` rather than reexported from `Base`?

tibbe: This is why this is tricky. If these functions take an existing value from the map, don't modify it, and stick it in a different map, then the value doesn't need to be evaluated (because the invariant is that values in a `Strict` map are already in WHNF). However, if the value comes from the user or is modified, then we need to evaluate it before inserting it.

tibbe: Staring at this I'm not even sure it's correct. We need to evaluate the expression `(f v w)` before inserting it into the map.

mschristiansen: If I rewrite `maybe` to `case f v w of` then `f v w` should be fully evaluated, right?

```
differenceWith :: (Eq k, Hashable k) => (v -> w -> Maybe v) -> HashMap k v -> HashMap k w -> HashMap k v
differenceWith f a b = foldlWithKey' go empty a
  where
    go m k v = case HM.lookup k b of
                 Nothing -> insert k v m
                 Just w  -> case f v w of
                             Just y -> insert k y m
                             Nothing -> m
```

mschristiansen: Suppose we still need to check that `y` is evaluated. `intersectionWith`, etc. would have same problem. `IntersectionWith` is doing a straight `insert k (f v w) m`.

mschristiansen: something like `Just !y -> insert k y m`?

mschristiansen: `BangPattern` is only one constructor deep, but using `deepseq` requires a typeclass constraint for `NFData`. Not sure what else to do.

tibbe:
```haskell
Just y -> insert k y m
```

is enough as long as we're using the strict `insert`. `foldlWithKey'` will evaluate the insert expression, which in turn always evaluates its second argument.

tibbe: In fact your original is right.

PR #156, opened 2017-06-13
Make Ord instance Eq compatible

phadej:


treeowl:
* I'm concerned that the test cases may not hit enough collisions to really exercise that aspect. And that aspect is the one that concerns me most. Are two equal maps built in different ways actually guaranteed to compare `EQ`?

* Shouldn't it be possible to make the `Ord` instance faster by coding it separately from the much wilder `Ord1` and `Ord2`?

* Would it make sense to offer more efficient, but more constrained, equality and comparison functions? `Ord` constraints on the keys and values seem to open up some options.

phadej:
> I'm concerned that the test cases may not hit enough collisions to really exercise that aspect. And that aspect is the one that concerns me most. Are two equal maps built in different ways actually guaranteed to compare EQ?

The model test should exercise that. I tested locally with bigger test-cases and test-sizes, still OK.

> Shouldn't it be possible to make the Ord instance faster by coding it separately from the much wilder Ord1 and Ord2?

Yes. `unorderedCompare` is slow, and `compare `on` sort` would most likely be faster. But again, is there a real-world example where collision sets are of non-trivial size?

/Side-note:/ we could have some `stats` function telling about that. It falls into same "debugging" category as `valid`.

> Would it make sense to offer more efficient, but more constrained, equality and comparison functions? Ord constraints on the keys and values seem to open up some options.

It's not written as a law, but I'd expect `compare2 ≡ compare`. So for different variants we'd need a `newtype`.

phadej:
Side-note: If you need to `compare` a lot of `HashMap`s, you'd probably better with `Map` anyway, do you agree? I made `Ord` because it's possible, not really motivated by real use cases. Though having `Ord Value` for `aeson` might turn out useful...

phadej:
Transivity... difficult. I have little understanding what it means when we have

- **If** [a] < [b] and [b] < [c] (where < is either `unorderedCompare cmpAB` or `unorderedCompare cmpBC`)
- **Then** [a] < [c], but what's `cmpAC :: a -> c -> Ordering` - it should have some properties related to `cmpAB` and `cmpBC`.

I think the sketch above is persuasive enough when
`cmpAB`, `cmpBC` and thus `cmpAC` are `coerce compare`.

But what if `cmp*` is comparing between equivalence classes?
I cannot break the algorithm below, neither the one in `unordered-containers`.
But I have no hard evidence that it works.

---

TL;DR we cannot fully `sort` input lists `as` or `bs`, but we can sort elements into *buckets*, so `cmp` cannot distinguish elements of a single bucket when compared to anything else in the other list. And these buckets are ordered.

One way is to rewrite cmpA and cmpB as in (which might be a good idea):

```haskell
unorderedCompare :: (Show a, Show b) => (a -> b -> Ordering) -> [a] -> [b] -> Ordering
unorderedCompare cmp as bs = go (sortBy cmpA as) (sortBy cmpB bs)
  where
    go [] [] = EQ
    go [] (_ : _) = LT
    go (_ : _) [] = GT
    go (x : xs) (y : ys) = cmp x y `mappend` go xs ys

    cmpA a a' = goB bs where
        goB [] = EQ
        goB (b : bs') = case (cmp a b, cmp a' b) of
            (LT, GT) -> LT
            (GT, LT) -> GT
            (LT, EQ) -> LT
            (GT, EQ) -> GT
            (EQ, LT) -> GT
            (EQ, GT) -> LT
            (EQ, EQ) -> EQ
            (LT, LT) -> goB bs'
            (GT, GT) -> goB bs'

    cmpB b b' = goA as where
        goA [] = EQ
        goA (a : as') = case (cmp a b, cmp a b') of
            (LT, GT) -> GT
            (GT, LT) -> LT
            (LT, EQ) -> GT
            (GT, EQ) -> LT
            (EQ, LT) -> LT
            (EQ, GT) -> GT
            (EQ, EQ) -> EQ
            (LT, LT) -> goA as'
            (GT, GT) -> goA as'
```

The first step is to sort input lists. When lists are "sorted" in compatible
way, the comparison itself is straightforward

**Sorting**. Let's dig into how cmpA (cmpB) work.

We try to find a relavant `b` in `bs` so we have 

- a < b && b < a'
- a < b && b <= a'
- ...

from which we can conclude by transitivity (and anti-symmetry) that a < a', a == a' or a > a' holds.

So what about the case where there aren't such points?  E.g.

```
[A1, A2] and [B5]?
[A2, A1] and [B5]
```

Well... if we think more broadly, we have found /some/ equivalence classes (= *buckets*).
Both examples above are 

```
[A_I,A_I] and [B_II]
```

so when we `go` through the elements in the *buckets* we it's irrelevant whether A1 or A2 comes
first, they are both less than `B5`, so algorithm works.
Even there are equivalence classes,
they are precise enough that `cmp` cannot distinguish between elements in them.

---

Actually (formally) *proving* that this is correct is non-trivial
(Agda or Coq and sorting makes my head hurt a little...).

In fact, I don't know how to even state the problem.
Proving reflexivity and anti-symmetry -like properties is trivial.

Transitivity...

---

The (QuickCheck) program I used:

```haskell
{-# LANGUAGE GeneralizedNewtypeDeriving #-}
import Data.List (sort,sortBy)
import Test.QuickCheck

import Debug.Trace

unorderedCompare :: (Show a, Show b) => (a -> b -> Ordering) -> [a] -> [b] -> Ordering
unorderedCompare cmp as bs = go (sortBy cmpA as) (sortBy cmpB bs)
  where
    go [] [] = EQ
    go [] (_ : _) = LT
    go (_ : _) [] = GT
    go (x : xs) (y : ys) = cmp x y `mappend` go xs ys

    cmpA a a' = compare (inB a) (inB a')
    cmpB b b' = compare (inA b) (inA b')

    inB a = (length $ filter (\b -> cmp a b == GT) bs, negate $ length $ filter (\b -> cmp a b == LT) bs)
    inA b = (length $ filter (\a -> cmp a b == LT) as, negate $ length $ filter (\a -> cmp a b == GT) as)

{-
    cmpA a a' = goB bs where
        goB [] = EQ
        goB (b : bs') = case (cmp a b, cmp a' b) of
            (LT, GT) -> LT
            (GT, LT) -> GT
            (LT, EQ) -> LT
            (GT, EQ) -> GT
            (EQ, LT) -> GT
            (EQ, GT) -> LT
            (EQ, EQ) -> EQ
            (LT, LT) -> goB bs'
            (GT, GT) -> goB bs'

    cmpB b b' = goA as where
        goA [] = EQ
        goA (a : as') = case (cmp a b, cmp a b') of
            (LT, GT) -> GT
            (GT, LT) -> LT
            (LT, EQ) -> GT
            (GT, EQ) -> LT
            (EQ, LT) -> LT
            (EQ, GT) -> GT
            (EQ, EQ) -> EQ
            (LT, LT) -> goA as'
            (GT, GT) -> goA as'
-}
```

### "coerce"

```haskell
data A = A1 | A2 | A3 | A4 | A5 deriving (Eq, Ord, Show, Enum, Bounded)
data B = B1 | B2 | B3 | B4 | B5 deriving (Eq, Ord, Show, Enum, Bounded)

instance Arbitrary A where
    arbitrary = arbitraryBoundedEnum
    shrink A1 = []
    shrink x  = [ minBound .. pred x ]

instance Arbitrary B where
    arbitrary = arbitraryBoundedEnum
    shrink B1 = []
    shrink x  = [ minBound .. pred x ]

aToB :: A -> B
aToB A1 = B1
aToB A2 = B2
aToB A3 = B3
aToB A4 = B4
aToB A5 = B5

eqProp :: [A] -> Bool
eqProp as = unorderedCompare (\a b -> aToB a `compare` b) as bs == EQ
  where
    bs = map aToB as

ordProp :: [A] -> [A] -> Bool -- Property
ordProp as as' =
    unorderedCompare (\a b -> aToB a `compare` b) as bs'
    ==
    compare (sort as) (sort as')
  where
    -- bs = map aToB as
    bs' = map aToB as'
```

### Transitivity example

```
newtype A = A Int deriving (Eq, Ord, Show, Arbitrary)
newtype B = B Int deriving (Eq, Ord, Show, Arbitrary)
newtype C = C Int deriving (Eq, Ord, Show, Arbitrary)

cmpAB :: A -> B -> Ordering
cmpAB (A a) (B b) = (a * 2) `compare` b

cmpBC :: B -> C -> Ordering
cmpBC (B b) (C c) = (b * 2) `compare` c

cmpAC :: A -> C -> Ordering
cmpAC (A a) (C c) = (a * 4) `compare` c

ordProp :: [A] -> [B] -> Property
ordProp as bs = lhs === rhs where
    lhs = unorderedCompare cmpAB as bs
    rhs = sort [ B (a * 2) | A a <- as ] `compare` sort bs

transProp :: [A] -> [B] -> [C] -> Property
transProp as bs cs = case (unorderedCompare cmpAB as bs, unorderedCompare cmpBC bs cs, unorderedCompare cmpAC as cs) of
    (EQ, EQ, ac) -> label "trans" $ ac == EQ
    (LT, LT, ac) -> label "trans" $ ac == LT
    (LT, EQ, ac) -> label "trans" $ ac == LT
    (EQ, LT, ac) -> label "trans" $ ac == LT
    (GT, GT, ac) -> label "trans" $ ac == GT
    (EQ, GT, ac) -> label "trans" $ ac == GT
    (GT, EQ, ac) -> label "trans" $ ac == GT
    -- dunno
    (GT, LT, _) -> label "gtlt" True 
    (LT, GT, _) -> label "gtlt" True 
```

treeowl:
I don't really understand your sketch too terribly well. I believe that in order to talk sensibly about transitivity (most especially to encode it into a proof system, but even informally), we have to assume that, to a very good approximation, `a ~ b`. We have, then,

```haskell
as, bs, cs :: [a]
```

The "unordered comparison" of `as` and `bs` will break each of them into equivalence classes, and the unordered comparison of `bs` and `cs` will break each of them into classes. Each of these classes may be *coarser* than the "natural" one for the comparison, and will be compatible with it. The million-dollar question is whether they're guaranteed to be compatible *with each other* in the appropriate sense. I don't yet see a reason (aside from randomized testing) to believe that to be the case.

treeowl:
The "modern" lifted comparison classes are such a horrible mess for sets and maps that I'm very sorely tempted to remove the `Eq2` and `Ord2` instances for `Map` and `HashMap`, and the `Eq1` and `Ord1` instances for `Set` and `HashSet`, until someone finds a way to make them sensible again. On the other hand, people will probably yell at me....

phadej:
I have a graphical version. I need to get color pens to draw it in understandable way though (I like pretty pictures). Give me few days

phadej:
Ok. I have worked on this a little, and made some progress. Unfortunately somewhat busy, but hopefully I'll find time to write things down tomorrow.

I'm afraid that `Ord2 Map` is bogus, but `Ord2 HashMap` actually is "correct".

treeowl:
When is even Eq2 for HashMap correct? Surely there must be a connection
between the Hashable instance and the equivalence...

phadej:
@treeowl I think I can answer that too. (haven't yet).

phadej:
```agda
module Equ where

open import Level
open import Relation.Binary.Core

-- Hetero-transitivity, or 2-hop / 4-element transitivity
--
--   A       B       A       B
--   x₂ <--> y₁ <--> x₁ <--> y₂
--
-- implies
--
--   x₂ <------------------> y₂
--
Law : ∀ {a b ℓ} {A : Set a} {B : Set b} → REL A B ℓ → Set _
Law _∼_ = ∀ {x₁ x₂ y₁ y₂}
        → x₁ ∼ y₁ → x₁ ∼ y₂ → x₂ ∼ y₁
        → x₂ ∼ y₂

-- 4-element transitivity implies ordinary transitivity: we make one hop reflexive.
Law→Transitive : ∀ {a ℓ} {A : Set a} → (_∼_ : Rel A ℓ)
  → Law _∼_
  → Reflexive _∼_
  → Symmetric _∼_
  → Transitive _∼_
Law→Transitive _ law r s = λ ij jk → law (s ij) jk r

-- In other direction, we use transitivity twice.
--
-- Reflexivity isn't used
IsEquivalence→Law : ∀ {a ℓ} {A : Set a} → (_∼_ : Rel A ℓ)
  → IsEquivalence _∼_
  → Law _∼_
IsEquivalence→Law _ record { refl = r ; sym = s ; trans = t } x₁y₁ x₁y₂ x₂y₁
  = t x₂y₁ (t (s x₁y₁) x₁y₂)

-- A homogeneous binary-relation we can construct from hetoregenous relation.
data R {a b ℓ} {A : Set a} {B : Set b} (r : REL A B ℓ) : A → A → Set (b ⊔ ℓ) where
  RRefl : ∀ {x} → R r x x
  RNorm : ∀ {x₁ x₂ y} → r x₁ y → r x₂ y → R r x₁ x₂

-- R _∼_ is equivalance if _~_ is 4-transitive
R-Reflexive : ∀ {a b ℓ} {A : Set a} {B : Set b} (r : REL A B ℓ) → Reflexive (R r)
R-Reflexive r = RRefl

R-Symmetric : ∀ {a b ℓ} {A : Set a} {B : Set b} (r : REL A B ℓ) → Symmetric (R r)
R-Symmetric r RRefl           = RRefl
R-Symmetric r (RNorm x₁y x₂y) = RNorm x₂y x₁y
  
R-Transitive : ∀ {a b ℓ} {A : Set a} {B : Set b} (r : REL A B ℓ) → Law r → Transitive (R r)
R-Transitive r law RRefl            RRefl           = RRefl
R-Transitive r law RRefl            (RNorm x₁y x₂y) = RNorm x₁y x₂y
R-Transitive r law (RNorm x₁y x₂y)  RRefl           = RNorm x₁y x₂y
R-Transitive r law (RNorm iy₁ jy₁ ) (RNorm jy₂ ky₂) = RNorm (law jy₁ jy₂ iy₁) ky₂

-- If we map elements of A to elements of B,
-- and we have an equivalance on B,
-- then we get induced equivalence on A.
F : ∀ {a b ℓ} {A : Set a} {B : Set b} (f : A → B) → (_∼_ : Rel B ℓ) → Rel A ℓ
F f _∼_ x y = f x ∼ f y

F-IsEquivalence
  : ∀ {a b ℓ} → {A : Set a} → {B : Set b} (f : A → B) → (_∼_ : Rel B ℓ)
  → IsEquivalence _∼_ → IsEquivalence (F f _∼_)
F-IsEquivalence f _∼_ record { refl = r ; sym = s ; trans = t } = record
  { refl  = λ {_} → r
  ; sym   = λ {_} {_} z → s z
  ; trans = λ {_} {_} {_} → t
  }

-- Similarly as with F-IsEquivalance;
-- If we map both sides, A and B, to C,
-- and we have an equivalence on C,
-- then we get induced 4-transitivity between A and B.
FG : ∀ {a b c ℓ}
   → ∀ {A : Set a} {B : Set b} {C : Set c}
   → (f : A → C) (g : B → C)
   → (_∼_ : Rel C ℓ)
   → REL A B ℓ
FG f g _∼_ x y = f x ∼ g y

FG-IsEquivalence-Law
  : ∀ {a b c ℓ}
  → ∀ {A : Set a} {B : Set b} {C : Set c}
  → (f : A → C) (g : B → C)
  → (_∼_ : Rel C ℓ)
  → IsEquivalence _∼_
  → Law (FG f g _∼_)
FG-IsEquivalence-Law f g _∼_ record { refl = r ; sym = s ; trans = t } x₁y₁ x₁y₂ x₂y₁
  = t x₂y₁ (t (s x₁y₁) x₁y₂)

-- How is this related to Eq1 HashSet in Haskell?
--
-- recall: class Eq1 f where liftEq :: (a -> b -> Bool) -> f a -> f b -> Bool
--
-- 1. I suggest that an argument to liftEq (of type a -> b -> Bool) should
--    witness some 4-transitive relation, _∼_
--
-- 2. Also importantly the R _∼_ : Rel A _ and R (flip _∼_) : Rel B _  should
--    contain Eq A and Eq B. Otherwise Hashable A or Hashable B might separate
--    equivalent element into different buckets.
--
-- 3. eq x y = True should imply hashWithSalt salt x = hashWithSalt salt y,
--    in the same say x1 == x2, implies hashWithSalt salt x1 = hashWithSalt salt x2
--
-- I conjecture, that a property like 1. makes unorderedCompare work "sensibly".
```

```haskell
{-# LANGUAGE CPP #-}
{-# LANGUAGE GeneralizedNewtypeDeriving #-}
{-# LANGUAGE ScopedTypeVariables #-}
import Data.Function (on)
import Data.List (partition)
import Debug.Trace
import Data.Functor.Classes (Eq1 (..))

#ifdef MIN_VERSION_unordered_containers
import Data.Hashable
import qualified Data.HashSet as HS
#endif

import Data.Ord (Down (..))
import qualified Data.Set as Set

-- | Treating lists as sets.
--
-- >>> unorderedEq (==) (3:[1..10]) (1:1:2:2:[1..10])
-- True 
--
-- >>> unorderedEq (==) (3:[1..10]) (1:1:2:2:[1..11])
-- False
--
-- >>> unorderedEq (on (==) (`mod` 3)) (3:[1..10]) (1:1:2:2:[1..10])
-- True
--
-- The correctness of algorithm depends on 4-transitivity.
-- Without it the 1. partition step won't find all elements?
--
-- For example:
--
-- >>> let dist n m = abs (n - m) <= (1 :: Integer)
--
-- is reflexive, and symmetric; but not transitive (transitivity implies 4-transitivity)
--
-- >>> (dist 0 1, dist 1 2, dist 0 2)
-- (True,True,False)
--
-- >>> unorderedEq dist [0,1,2] [1]
-- True
--
-- >>> unorderedEq dist [0,1,2] [1,3]
-- False
--
-- >>> unorderedEq dist [2,1,0] [1,3]
-- True
--
-- Note: Show context is so you can add traceShow where needed :)
--
-- Note: isPermutationBy in unordered-containers treats lists as multisets.
--
unorderedEq :: forall a b. (Show a, Show b) => (a -> b -> Bool) -> [a] -> [b] -> Bool
unorderedEq eq = go where
  go []       []    = True
  go []       (_:_) = False
  go (x : xs) ys
      -- 2. if xys are empty: there are elements in xs0 which aren't in ys.
      | null xys = False
      -- 3. otherwise we have to remove from the xs all elements which are equal to any element of xys.
      | otherwise = go (filter (\x -> not $ elemBy eq x xys) xs) yys
    where
      -- 1. xys are elements of ys which are equal to x
      xys, yys :: [b]
      (xys, yys) = partition (eq x) ys

-- Note: heterogenity
elemBy :: (a -> b -> Bool) -> a -> [b] -> Bool
elemBy eq a = go where
  go [] = False
  go (b : bs) = eq a b || go bs

-------------------------------------------------------------------------------
-- broken: HashSet
-------------------------------------------------------------------------------

#ifdef MIN_VERSION_unordered_containers
newtype D = D Int deriving (Eq, Show, Hashable)

lhs :: [D]
lhs = map D [1..3]

rhs :: [Int]
rhs = [2,4,6]

-- This is 4-transitive
dIntEq :: D -> Int -> Bool
dIntEq (D n) m = n * 2 == m

ex1a = unorderedEq dIntEq lhs rhs -- True
ex1b = liftEq dIntEq (HS.fromList lhs) (HS.fromList rhs) -- False !!! wrong

#endif

-------------------------------------------------------------------------------
-- broken: Set
-------------------------------------------------------------------------------

lhs2 :: Set.Set Int
lhs2 = Set.fromList [1..10]

rhs2 :: Set.Set (Down Int)
rhs2 = Set.fromList $ map Down [1..10]

ex2 = liftEq (\n (Down m) -> n == m) lhs2 rhs2 -- False !!! wrong

-------------------------------------------------------------------------------
-- Main for ...
-------------------------------------------------------------------------------

main :: IO ()
main = pure ()
```

---

TL;DR, the `Eq1 HashSet` (and other `Eq1/Ord1/Eq2/Ord2` 'Set/HashSet/Map/HashMap`) work, but the constraints required for the argument equality or comparison are quite strict.

I'm now personally convinced that adding these instances wasn't the best move.

If you, @treeowl going to remove the instances, I'd ask you to consider adding functions like

```haskell
-- safe if Eq/Hashable a/b agree (e.g. newtype wrapper over `Text` to avoid stringly programming).
semiSafeCoerce  :: Coercible a b => HashSet a - > HashSet b
```

which would reduce the need to compare heterogenously. The requirement on `semiSafeCoerce` is similar which would make `liftEq` behave correctly. Yet, `semiSafeCoerce` and alike would be more generally useful.

PR #157, opened 2017-06-13
Generating expr

### LINKED ISSUE #154: Arbitrary instance for HashMap is unfortunate
`arbitrary` for `HashMap k v` (in `tests/Strictness.hs`) is based on `fromList`. This means that only `HashMap` shapes that can be produced by `fromList` will be represented in the inputs to test functions.

phadej:
This is try to fix #154 

We could rewrite all `[Key]` or `[(Key, Int)]` based properties into `Expr Key ()` or `Expr Key Int` based ones, respectively. I think including `union` is worth it, as it is has non-trivial implementation.

First commit is from #156.

### DIFF
--- a/tests/Utils.hs
+++ b/tests/Utils.hs
@@ -1,0 +1,105 @@
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+-- | Test utilities
+module Utils
+    ( Key (..)
+    , Expr (..)
+    , evalExprStrict
+    , evalExprLazy
+    , evalExprSet
+    , evalExprOrdSet
+    ) where
+
+import Control.Applicative (liftA2, liftA3)
+import Data.Hashable (Hashable(hashWithSalt))
+import Test.QuickCheck (Arbitrary(arbitrary,shrink), oneof, sized)
+
+import qualified Data.HashMap.Strict as HMS
+import qualified Data.HashMap.Lazy as HML
+import qualified Data.HashSet as HS
+import qualified Data.Set as Set
+
+-- | A key with collision-prone Hashable instance
+------------------------------------------------------------------------
+
+-- Key type that generates more hash collisions.
+newtype Key = K { unK :: Int }
+   deriving (Arbitrary, Enum, Eq, Ord, Read, Show, Integral, Num, Real)
+
+instance Hashable Key where
+    hashWithSalt salt k = hashWithSalt salt (unK k) `mod` 20
+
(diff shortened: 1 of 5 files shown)


tests/Utils.hs
```diff
@@ -0,0 +1,105 @@
...
+  deriving Show
+
+instance (Arbitrary k, Arbitrary v) => Arbitrary (Expr k v) where
+    arbitrary = sized arb
+      where
+        arb n | n <= 0 = oneof leafs
+        arb n = oneof $ leafs ++
+            [ liftA3 ExprInsert arbitrary arbitrary (arb (n - 1))
+            , liftA2 ExprDelete arbitrary (arb (n - 1))
+            , liftA2 ExprUnion (arb (n `div` 2)) (arb (n - n `div` 2))
```
treeowl: Why `div` and not `quot`?

phadej:
```
Prelude Test.QuickCheck> 3 `div` 2 + 3 `quot` 2
2
```

phadej:
> I'm not quite convinced this fully solves the problem. We want to guarantee (as best we can) that every valid HashMap tree is available to the Arbitrary instance. Do we have good reasons to believe that all such trees can be generated by this language?

I don't know, I looked at non-trivial `-> HashMap k v` functions in ` Base.hs` module.

> An alternative, more complicated, approach (but quite possibly the right one) would be to use an expression language that includes all primitive operations, interpreting it using HashMap or HashSet and also (say) Data.Map or Data.Set.

AFAICS the `insert`, `delete` and `union` are a good start.

> One of the key ideas of the containers test suite is that all the invariants for each data structure are expressed in a single valid function. In this context, that would allow your Expr interpretation to find the smallest invalid subexpression, rather than waiting for that invariant failure to show up as a hard-to-debug semantic error.

That's a good idea, but it can be added independently. I'm not 100% sure what are the invariants.

treeowl:
Can you rebase this, please?

phadej:
rebased

phadej:
ping @treeowl, is there something I can do to get this off my "open PR" -list? :)

treeowl:
Uh, I seem to have forgotten about this. I'll try to get to it in the next couple days.

phadej:
This seems to be still forgotten...

treeowl:
I think we've removed the need for it by removing the representation
redundancy. See the changes in Eq implementation.

PR #163, opened 2017-08-07
Add memory footprint measurement script (renamed from "Unpack leaves")

ethercrow:
Memory footprint of `HashMap Int Int` according to ghc-datasize:

```
Size  | Bytes before | Bytes after
   0  |            8 |           8
   1  |          112 |         112
   2  |          264 |         232
   4  |          440 |         376
   8  |          792 |         664
  16  |         1472 |        1216
  32  |         4656 |        4144
  64  |         7984 |        6960
 128  |        14640 |       12592
 256  |        27568 |       23472
 512  |        78768 |       70576
1024  |       132016 |      115632
2048  |       238512 |      205744
```

Runtimes in `stack bench` seem to be the same with the patch.

### DIFF
--- a/tests/MemoryFootprint.hs
+++ b/tests/MemoryFootprint.hs
@@ -1,0 +1,46 @@
+{-# language BangPatterns #-}
+
+import Control.Monad
+import qualified Data.HashMap.Strict as HM
+import qualified Data.Map.Strict as M
+import qualified Data.IntMap as IM
+import GHC.DataSize
+import Text.Printf
+import Control.DeepSeq
+
+sizes :: [Int]
+sizes = [2^n | n <- [1 :: Int .. 11]]
+
+main :: IO ()
+main = do
+    p "(Int, Int)" (1 :: Int, 2 :: Int)
+    putStrLn ""
+    p "HM Int Int (0)" (HM.empty :: HM.HashMap Int Int)
+    p "HM Int Int (1)" (HM.singleton (1 :: Int, 42 :: Int))
+    forM_ sizes $ \n -> pn "HM Int Int" n (HM.fromList (zip [0 :: Int .. (n - 1)] [0 :: Int ..]))
+    putStrLn ""
+    p "IM Int (0)" (IM.empty :: IM.IntMap Int)
+    p "IM Int (1)" (IM.singleton (1 :: Int) (42 :: Int))
+    forM_ sizes $ \n -> pn "IM Int" n (IM.fromList (zip [0 :: Int .. (n - 1)] [0 :: Int ..]))
+
+    putStrLn ""
+    p "(String, Int)" ("1", 2 :: Int)
+    putStrLn ""
+    p "HM String Int (0)" (HM.empty :: HM.HashMap String Int)
+    p "HM String Int (1)" (HM.singleton ("1", 42 :: Int))
+    forM_ sizes $ \n -> pn "HM String Int" n (HM.fromList (zip (map show [0 :: Int .. (n - 1)]) [0 :: Int ..]))
+    putStrLn ""
+    p "M String Int (0)" (M.empty :: M.Map String Int)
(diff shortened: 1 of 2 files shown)


treeowl:
A more difficult change, whose effects are substantially harder for me to guess, would be to try splitting the arrays of `HashMap`s into arrays of keys, arrays of values, and arrays of non-trivial `HashMap`s. This would make nodes a good bit heavier and eliminate the `Full` optimization, but I imagine it could reduce indirection quite a bit in some cases. You may or may not want to experiment with that idea. In any case, please try out the `UNPACK` simplification of your current idea and let me know.

treeowl:
There is another potential optimization, of `HashSet`, that this would interfere with. I don't know if it's worth worrying about, though. In particular, we *could* use

```haskell
newtype Leaf k f = L (f k)
data SSP v k = SSP !k v
data HashMap' k (f :: * -> *) =
   ...
   | Leaf !(Leaf k f)
newtype HashMap k v = HashMap (HashMap' k (SSP v))
newtype HashSet k = HashSet (HashMap' k Identity)
```

This would remove the useless `()` slots from `HashSet`, but at the expense of losing the ability to unpack leaves into the `Leaf` constructor. As I said, I suspect we should go with the leaf unpacking as you suggest, but it's something to ponder.

ethercrow:
After adding the measurement script as a test suite to unordered-containers, I'm getting the same numbers for master, my initial change and `{-# UNPACK #-}` approach. Looks like I was just comparing to a different version of the package. Oh well.

Let's use this pull request to add the measurement script then, so that some people from the future don't do the same mistake.

treeowl:
That doesn't make any sense to me. Your change, and the `UNPACK` version of it, should surely reduce the in-memory size of each `HashMap` by at least a couple words per key. That's supported not only by common sense but also by reading the Core. So I'm not buying this "They're all the same" conclusion.

ethercrow:
> A more difficult change, whose effects are substantially harder for me to guess, would be to try splitting the arrays of HashMaps into arrays of keys, arrays of values, and arrays of non-trivial HashMaps. This would make nodes a good bit heavier and eliminate the Full optimization, but I imagine it could reduce indirection quite a bit in some cases. You may or may not want to experiment with that idea.

Yes, I wanted to experiment with that after reading about CHAMP hashmaps.

ethercrow:
Running the measurement with `+RTS -s` also yields essentially the same numbers of bytes allocated for both master and unpacked version.

treeowl:
I'll take a look tomorrow if I have time. I'm guessing you're not
rebuilding properly, but maybe I'm missing something.

tibbe:
Unpacking the leaf makes sense to me and I don't know why I didn't do it (`Leaf` wasn't in the original design so I might have forgotten an unpack when I added it). We should make sure that we can actually measure and improvement though. It ought to reduce memory usage so I'm surprised if it doesn't.

ethercrow:
Just checking that ghc-datasize properly reports sizes when unpacking is involved:
```
data Bar a b = Bar !a b
data Foo a b c = Foo !a !(Bar b c)
data Foo' a b c = Foo' !a {-# UNPACK #-} !(Bar b c)
data Foobar a b c = Foobar !a !b c

Bar Int Int    56
Foo Int Int Int     96
Foo' Int Int Int   80
Foobar Int Int Int 80
```

So at least this seems to make sense.

ethercrow:
I've just tried this: `| Leaf !Hash {-# NOUNPACK #-} !(Leaf k v)` and the footprint numbers actually go up! Looks like GHC is unpacking the leaves on its own, but I have no idea why, I thought it does that only when fields are pointer-sized.

ethercrow:
Actually, it's now obvious:
```
{-# OPTIONS_GHC -fno-full-laziness -funbox-strict-fields #-}
```

tibbe:
Ah. That's where I put it. I looked in the .cabal file for a program earlier today but didn't see it there and got temporarily confused.

Issue #126, opened 2016-02-21
Using SmallArray#


AndrasKovacs:
It's been a while since Johan added SmallArray# to GHC. Is there a reason to not use it here? Searching this repository didn't bring up relevant discussion. I did some benchmarking and it overall it showed improvement  (here's a [data dump](http://lpaste.net/153003)).

tibbe:
Makes sense. We need to support the last three major versions of GHC, so
you might need to add some ifdefs in the Array module.

bgamari:
I believe `SmallArray` was added in 7.10, so in principle support could be added after the GHC 8.2 release without the `ifdefs` (although retaining compatibility would probably be worthwhile). @AndrasKovacs, do you have a patch for this?

treeowl:
Closed by #164.

PR #164, opened 2017-08-08
Use SmallArray#

ethercrow:
As discussed here https://github.com/tibbe/unordered-containers/issues/126 it's now appropriate time for this PR.


ethercrow:
Test suite fails to build with 8.2.

ethercrow:
@nad Could you allow syb-0.7 in ChasingBottoms please?

treeowl:
Ah, good point. I'll review as soon as I can.

treeowl:
Do we really need to drop support for 7.8? How bad would it be to add a compat module with type synonyms and wrapper functions? The `SmallArray#` interface looks pretty much the same as the `Array#` one aside from names. I know we're not *obligated* to continue supporting 7.8, but I think we should only drop support for a GHC version when it becomes a substantial burden.

ethercrow:
How about this way?

ethercrow:
OK, this should be better.

treeowl:
Sorry, I may not have been clear enough. I *do* think it makes good sense to drop support for super-old versions. I personally feel pretty comfortable dropping support for everything before 7.8 right now. To support 7.8, I think there should probably be a separate `SmallArray` module, so we only need to change the import line in `Array`.

ethercrow:
OK dropped everything older than 7.8

Data/HashMap/Array.hs
```diff
@@ -78,6 +81,24 @@ import qualified Prelude
 
 import Data.HashMap.Unsafe (runST)
 
+
+#if __GLASGOW_HASKELL__ >= 710
+type Array# a = SmallArray# a
+type MutableArray# a = SmallMutableArray# a
```
treeowl: You should probably use

```haskell
type Array# = SmallArray#
type MutableArray# = SmallMutableArray#
```

Not a big deal.

Data/HashMap/Base.hs
```diff
@@ -1375,18 +1345,12 @@ fullNodeMask = complement (complement 0 `unsafeShiftL` maxChildren)
...
 
-#if __GLASGOW_HASKELL__ >= 708
 ------------------------------------------------------------------------
 -- IsList instance
 instance (Eq k, Hashable k) => Exts.IsList (HashMap k v) where
     type Item (HashMap k v) = (k, v)
     fromList = fromList
-    toList   = toList
-#endif
+    toList   = toList
```
treeowl: Please put in the missing newline.

treeowl:
One more thing: please put the benchmark results in the commit message.

treeowl:
Oh, blast! Did GHC get the type role wrong for `SmallArray#`???

ethercrow:
And I'm getting the same error with GHC 8.2

treeowl:
Oh, no, I was just wrong. You need the type synonyms to take arguments. Sorry. Dunno why...

ethercrow:
Oh, I see

treeowl:
I just opened https://ghc.haskell.org/trac/ghc/ticket/14101 to try to get this fixed. I can't imagine it's really supposed to work like that.

ethercrow:
I added bench runs to the main commit. I didn't fix CPU frequency or isolated cores or anything like that, so the results are not too trustworthy.

ethercrow:
Also funny and sad thing is, ghc-datasize seems to break on `SmallArray#`, so I didn't measure memory footprint.

treeowl:
Ugh. Benchmarks that run in *microseconds*? The improvements are small, but so are the (maybe?) regressions. And it's pretty clear that card marking won't do a 16-element rarely-mutated array any good, so let's do this.

treeowl:
Thanks, @ethercrow.

ethercrow:
I forgot to delete now unused cbits directory.

andrewthad:
I'm glad this has been merged, but I'd like to point out that users will no longer be able to add a hashmap to a compact region because the [SmallArray compaction code is not implemented yet](https://ghc.haskell.org/trac/ghc/ticket/13857).

PR #165, opened 2017-08-09
Use SmallArray# primops

ethercrow:
I've tried plugging primops from https://github.com/AndrasKovacs/array-primops into `Array.insert` and `Array.delete` and seen about 5% improvement in respective benchmarks.

### DIFF
--- a/Data/HashMap/Array.hs
+++ b/Data/HashMap/Array.hs
@@ -69,7 +69,9 @@ import GHC.Exts (SmallArray#, newSmallArray#, readSmallArray#, writeSmallArray#,
                  SmallMutableArray#, sizeofSmallArray#, copySmallArray#, thawSmallArray#,
                  sizeofSmallMutableArray#, copySmallMutableArray#)
 
+#if defined(ARRAY_PRIMOPS)
 import GHC.Prim.SmallArray
+#endif
 
 #else
 import GHC.Exts (Array#, newArray#, readArray#, writeArray#,
@@ -262,7 +264,7 @@ copyM !src !_sidx@(I# sidx#) !dst !_didx@(I# didx#) _n@(I# n#) =
 -- | /O(n)/ Insert an element at the given position in this array,
 -- increasing its size by one.
 insert :: Array e -> Int -> e -> Array e
-#if __GLASGOW_HASKELL__ >= 710
+#if __GLASGOW_HASKELL__ >= 710 && defined(ARRAY_PRIMOPS)
 insert ary@(Array a) idx@(I# i) e =
     CHECK_BOUNDS("insert", count + 1, idx)
     Array (insertSmallArray# i e a)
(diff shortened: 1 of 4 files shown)


treeowl:
Interesting.... It seems he gets away with this by making sure the garbage collector won't run between the time the array is allocated and the time it's filled with meaningful information. Cool idea. There are a couple of potential issues, however:

1. The `array-primops` package is not exactly popular. It doesn't even show up in the Hackage reverse dependency listing! So if we decide to use it, I want to make sure that it's easy to *stop* using it if it stops working and doesn't get fixed.

2. Since it's based on CMM, I don't think it will work with GHCJS or Haste. Perhaps `array-primops` could incorporate the conditional compilation necessary to support them; I really don't know. We'd have to ask @AndrasKovacs.

ethercrow:
The dependency can be under `if (impl(ghcjs))`.

AndrasKovacs:
FYI, there is a very unsafe option which is faster than `array-primops`. Since `array-primops` is cmm, it's never inline. But you can almost rewrite it in standard inline primops, by allocating memory with `newByteArray#`, then copying data (including info the pointer). It should be the case that this is not interruptible by GC, but obviously this is a rather hairy dependency on GHC implementation.

You can see it [here](https://github.com/AndrasKovacs/array-primops/blob/master/benchmarks/Benchmarks.hs#L340).  I got the following times for inserting into arrays of size ten:

```
benchmarking insert/SmallArray/10/prim
time                 24.93 ns   (24.92 ns .. 24.95 ns)

benchmarking insert/SmallArray/10/slow
time                 39.91 ns   (39.85 ns .. 39.98 ns)

benchmarking insert/SmallArray/10/unsafe
time                 18.24 ns   (18.22 ns .. 18.26 ns)
```

Note that we would need to write it slightly differently when profiling is on; in that case IIRC there are two extra words before objects which would need to be copied as well.

treeowl:
Is there no way to tell GHC to inline the cmm? If not, well, that could be
a great thing to fix! Inlining the Haskell sounds ... pretty risky.

AndrasKovacs:
There's no way to inline them. It's been suggested e. g. [here](https://ghc.haskell.org/trac/ghc/wiki/Commentary/PrimOps#Foreignout-of-linePrimOpsandforeignimportprim) before.

ethercrow:
This is no longer WIP. The dependency is added only when SmallArray is used and compiler is not ghcjs. However, after running the benchmarks a few times, I get the improvement within the noise, so I don't think all this ifdefing is worth it.

ethercrow:
I'll give this PR some time to stay open just in case someone comes over and points out what I did wrong.

ethercrow:
https://github.com/bos/criterion/issues/115 this basically says that a single run showing 5% improvement means nothing.

PR #166, opened 2017-08-10
Remove now unused cbits

ethercrow:


treeowl:
Thanks, @ethercrow. Also thanks @andrewthad for bumping this.

PR #167, opened 2017-08-26
Add Data.UnorderedIntMap

ethercrow:
The code is just HashMap.Base with everything related to hashes deleted and keys specialized to Int. There are still some copy-paste artifacts, I'll fix those if there's interest.


ethercrow:
Criterion screenshot (github doesn't seem to want it in a comment): https://i.imgur.com/TswSzWA.png

ethercrow:
OK, that's understandable.

PR #168, opened 2017-09-09
Fix the build post-Semigroup–Monoid Proposal

### LINKED ISSUE #171: Update imports to be compatible with GHC-8.3 and higher Semigroup change.
The code seems to be already compatible.

RyanGlScott:
After the `Semigroup`–`Monoid` Proposal (now implemented in GHC HEAD), `Data.Semigroup` no longer exports `Monoid(..)`, causing `unordered-containers` to fail to build, since it tries to import `Monoid(..)` from `Data.Semigroup`. But we don't even need to do so, since `Monoid(..)` is already in scope, so we can delete this import safely.

Fixes #171


treeowl:
Do you think you could add GHC 8.2 (and maybe HEAD) to the CI?

treeowl:
Let me take that back; we *definitely* want HEAD.

treeowl:
Thanks. Unfortunately, it seems we have a missing `Semigroup` instance!

RyanGlScott:
Indeed, I'm working on fixing `primitive` to accommodate the SMP at the moment.

treeowl:
What's primitive have to do with it? I don't think we use that here.

RyanGlScott:
The build failure is due to `primitive` failing to install (it's a dependency of the test suite).

bgamari:
What is the status of this?

RyanGlScott:
I'm not sure. There still isn't a new release of `primitive` on Hackage that supports GHC HEAD, so in that regard nothing has changed. But perhaps our way of testing GHC HEAD shouldn't depend on Hackage, but rather an overlay like [`head.hackage`](https://github.com/hvr/head.hackage) (which already has a patch for `primitive`).

@hvr, do you know if there is a Travis CI script somewhere that takes advantage of `head.hackage` for its GHC HEAD builds?

asr:
Blocking https://github.com/agda/agda/issues/2878.

hvr:
@RyanGlScott 

> do you know if there is a Travis CI script somewhere that takes advantage of head.hackage for its GHC HEAD builds?

Yes, there is! @phadej recently added support to the travis script generator!

treeowl:
@hvr obviously, we can fix the `unordered-containers` CI build this way, but that won't stop the complaints from downstream; they'll *also* need to switch to Hackage head to get their builds working. *Nobody* can make a normal Hackage release for 8.4 until *all* their dependencies have done so. The domino effect means that we may have to wait for at least a few weeks *after* GHC makes a non-alpha release before Hackage can be brought properly up to date. If some maintainers happen to be off skiing in the mountains during this time, that could extend to months.

treeowl:
@hvr One option might be to designate git tags as dependencies, allowing a package to be *uploaded* to Hackage before the dependencies have been released. Then the packages would actually be *released* on Hackage once all the dependencies (with those exact tags) have been released. But we don't have any of that infrastructure yet.

RyanGlScott:
`primtive` status notwithstanding, I've updated the PR to use a `.travis.yml` file using `head.hackage`, so we should be able to verify that `unordered-containers` builds on GHC 8.4 and HEAD. Is there anything else that needs to be done here?

treeowl:
@RyanGlScott Looking at the Travis results, it *looks* like head is in the must-pass and 8.4 is in the may-fail. Am I reading that right? If so, that's not what we want. But that could just be a display thing.

RyanGlScott:
Sorry, that was a thinko I made in https://github.com/tibbe/unordered-containers/pull/168/commits/7dd8eaac2770f5a65dbc3bd4a7887fd0f4c0076e. That should be fixed as of the latest commit, https://github.com/tibbe/unordered-containers/pull/168/commits/32f0aed6c59f67cff0a2d8ccdb9f2a3164e1f218.

treeowl:
Oh, you just changed that. Nice.

treeowl:
@RyanGlScott , please squash into however many commits you want (the last two at least should be unified). Then I'll merge.

RyanGlScott:
I've squashed the commits, and the Travis build is green.

treeowl:
Thanks, @RyanGlScott

utils/Stats.hs
```diff
@@ -15,16 +16,21 @@ data Histogram = H {
...
           empty         = empty h1 + empty h2
         , leaf          = leaf h1 + leaf h2
         , bitmapIndexed = bitmapIndexed h1 + bitmapIndexed h2
         , full          = full h1 + full h2
         , collision     = collision h1 + collision h2
         }
 
+instance Monoid Histogram where
+    mempty = H 0 0 0 0 0
+#if __GLASGOW_HASKELL__ < 803
```
phadej: shouldn't this been `MIN_VERSION_base` check

RyanGlScott: Indeed. I prioritized maintaining the existing code conventions over my desire for a slightly more correct macro :)

Of course, that's not to say we shouldn't switch over to `MIN_VERSION_base` at some point.

hvr: `MIN_VERSION_base` would surely be less error-prone (and statically checked for form!), and the original reason for using `__GLASGOW_HASKELL__` is becoming more and more moot.

asr:
Since we are the subject, I understand the reasons for using `MIN_VERSION_base` instead of  `__GLASGOW_HASKELL__` on the above example, but what should I use when using CPP on the `Prelude`?

For example, for hiding `Prelude.<>` I can use

```
#if MIN_VERSION_base(4,11,0)
import Prelude hiding ((<>))
#endif
```

or

```
#if __GLASGOW_HASKELL__ >= 804
import Prelude hiding ((<>))
#endif
```
 
I think I should use `__GLASGOW_HASKELL__ ` because the `Prelude` is defined by GHC, but I would to know your opinions about it.

utils/Stats.hs
asr: Since we are the subject, I understand the reasons for using `MIN_VERSION_base` instead of  `__GLASGOW_HASKELL__` on the above example, but what should I use when using CPP on the `Prelude`?

For example, for hiding `Prelude.<>` I can use

```
#if MIN_VERSION_base(4,11,0)
import Prelude hiding ((<>))
#endif
```

or

```
#if __GLASGOW_HASKELL__ >= 804
import Prelude hiding ((<>))
#endif
```
 
I think I should use `__GLASGOW_HASKELL__ ` because the `Prelude` is defined by GHC, but I would to know your opinions about it.

phadej: `Prelude` is defined by `base`. AMP, FTP and now SMP are all library changes, and should be guarded with `MIN_VERSION_base`.

From top of my head, you want to use `__GLASGOW_HASKELL__` around specific `OPTIONS_GHC` or language pragmas, like `GeneralizedNewtypeDeriving`.

Sometimes you *should* guard on both, e.g. with `DeriveGeneric`, as you need both compiler (GHC) and library (base) support. But in practice guarding on GHC is enough, as "if GHC supports `DeriveGeneric`, there is `GHC.Generics` in `base`".

asr: Crystal. Thanks!

PR #175, opened 2018-01-23
Add operator HashMap.!?

m-renaud:
This is equivalent to `lookup` with the arguments flipped. This operator was
[added to containers-0.5.9.1](https://hackage.haskell.org/package/containers-0.5.10.2/docs/Data-Map-Strict.html#v:-33--63-), this makes the `HashMap` interface consistent.


treeowl:
This sounds reasonable.

treeowl:
You have legit build failures. You'll need to fix some import/export business.

m-renaud:
I wasn't sure if it would be beneficial to add benchmarks for this or not since its equivalent to `lookup` which already has benchmarks. WDYT?

treeowl:
Don't bother.

m-renaud:
Gotcha, this is ready for review.

m-renaud:
Resolved merge conflicts, PTAL!

Data/HashMap/Base.hs
```diff
@@ -623,6 +624,17 @@ lookupCont absent present !h0 !k0 !m0 = go h0 k0 0 m0
         | otherwise = absent (# #)
 {-# INLINE lookupCont #-}
 
+-- | /O(log n)/ Return the value to which the specified key is mapped,
+-- or 'Nothing' if this map contains no mapping for the key.
+--
+-- This is a flipped version of 'lookup'.
+--
+-- @since 0.2.9
```
treeowl: We sure about that number?

m-renaud: Out of date since last update, was just waiting for the findWithDefault fix to go into master since the CI was failing on this PR too. Have updated to 0.2.11 and updated CHANGES.md appropriately as well. Most recent commit should have the updated annotation.

Data/HashMap/Strict/Base.hs
```diff
@@ -39,6 +39,7 @@ module Data.HashMap.Strict.Base
     , size
     , HM.member
     , HM.lookup
+    , (HM.!?)
```
emilypi, line 42: Minor optional nitpick: this is a bit of an odd prefix to stare at in source. Can we just export `(!?)`?

m-renaud: So the operator `!?` is just being re-exported from the `Data.HashMap.Base` module which is imported qualified as `HM` (see line R105 below). Following convention here, all re-exported symbols without modification/strict-specific implementations for the `Strict` version are explicitly qualified with `HM.`. We _could_ `import Data.HashMap.Base ((!?))` which would let us just export `(!?)` in the list here, but then that could be misinterpreted as having a custom `!?` operator for the strict module.

emilypi: Makes sense, thanks.

Data/HashMap/Base.hs
```diff
@@ -623,6 +624,17 @@ lookupCont absent present !h0 !k0 !m0 = go h0 k0 0 m0
         | otherwise = absent (# #)
 {-# INLINE lookupCont #-}
 
+-- | /O(log n)/ Return the value to which the specified key is mapped,
+-- or 'Nothing' if this map contains no mapping for the key.
```
treeowl, line 628: We should probably revisit the `O(log n)` claim at some point, since our hashing is not too random for, e.g., `Int`. But that doesn't have to be handled in this PR.

PR #178, opened 2018-01-24
Add HashMap.adjustWithKey.

m-renaud:
This function is present in the container's Map interface, this brings more
consistency to the HashMap interface so that it can more easily be used as a
drop in replacement for Data.Map if desired.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@
  * Remove support for GHC versions before 7.8. (Thanks, Dmitry Ivanov!)
  * Use `SmallArray#` instead of `Array#` for GHC versions 7.10 and above.
    (Thanks, Dmitry Ivanov!)
+ * Add `HashMap.adjustWithKey`.
 
 ## 0.2.8.0
 
--- a/Data/HashMap/Base.hs
+++ b/Data/HashMap/Base.hs
@@ -26,6 +26,7 @@ module Data.HashMap.Base
     , unsafeInsert
     , delete
     , adjust
+    , adjustWithKey
     , update
     , alter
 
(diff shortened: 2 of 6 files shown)


treeowl:
Let's start with the most general: `alterF`. We can think about others later. I suspect the implementation should be similar to the one for `IntMap`, but you should be sure to calculate the hash only once.

m-renaud:
Sounds good, it looks like I'm going to have to pull out the `go` functions for `lookup`, `insert` and `delete` into top level helpers (so that I can pass the computed hash value into them instead of them computing it each time). Does that sound reasonable?

treeowl:
Sounds likely!

treeowl:
You'll probably need to do a bit more than that, actually. You want to
avoid testing key equality more than once. I think that means you'll need a
version of lookup that records the key position in case of collision, and
specialized versions of insert and delete to support the following
situations:

* insert, knowing that the key is not yet in the map
* insert, knowing that the key is in the map without collision
* insert, knowing the key is in the map with a particular position in a
collision
* delete, knowing the key is in the map without collision
* delete, knowing the key is in the map in a particular collision location.

You should be able to merge the functions for present keys together,
passing a bogus (negative) index when there was no collision.

treeowl:
Also, we're not adding adjustWithKey. I'd much rather *remove* it from
containers. It's a fussy micro-optimization and I just don't buy it.

m-renaud:
> we're not adding adjustWithKey

SGTM.

> version of lookup that records the key position in case of collision

Sounds reasonable.

> insert, knowing that the key is not yet in the map

I don't see how these would allow us to cut out any operations, you'd still need to traverse through the layers of `BitmapIndexed` and `Full` until you reach an `Empty`. Am I missing something?

> insert, knowing that the key is in the map without collision

I can see this, you can skip the `ky == k` check of the `Leaf` case.

> insert, knowing the key is in the map with a particular position in a collision

Yup, makes sense, just jump to the right index in the collision array

> delete, knowing the key is in the map without collision

If I'm reading the implementation right, if we know that a key `k` is in the map without collision then we will reach a `Leaf` constructor once, and that leaf **must** contain the key that we're deleting. Is that correct?

> delete, knowing the key is in the map in a particular collision location.

Yup, same as for insert, just jump to the right index in the collision array

treeowl:
When you insert knowing the key is not in the map, you don't have to test
equality if you find that its *hash* is in the map. These versions of
insert and delete won't need Eq constraints.

m-renaud:
So. Much. Code. Duplication. :P

treeowl:
Be grateful it's easier than the Data.Map version!

treeowl:
Not adding `adjustWithKey`; `alterF` was added in #181.

PR #180, opened 2018-01-26
Add HashMap.alterF.

m-renaud:
The implementation utilizes specialized versions of lookup, insert, and delete
to guarantee that the hash is only computed once and key comparisons are done at
most once.

**Testing:**

- Property tests are added to verify that the behaviour is equivalent to that of
  `Data.Map.alterF`.
- Property tests for `alter` as well since they weren't present.

**Benchmarking**

Benchmarks added for `alter` and `alterF`. Performance is comparable to `insert` and `delete` operatiors. See https://github.com/tibbe/unordered-containers/pull/180/commits/b143063c45868fce5eb1d4275119bcf37e8b0cf3 for details.

**Open questions/follow-ups:**

- Verify that strictness is enforced properly in Data.HashMap.Strict.alterF.
- Make another pass to determine if more checks can be removed from optimized
  insert and delete operations.

This partially addresses #172.

### DIFF
--- a/Data/HashMap/Base.hs
+++ b/Data/HashMap/Base.hs
@@ -468,9 +468,7 @@ lookup k0 m0 = go h0 k0 0 m0
 --   Key not in map           => Absent
 --   Key in map, no collision => Alone v
 --   Key in map, collision    => Collide v position
-lookupRecordCollision
-  :: (Eq k, Hashable k)
-  => Hash -> k -> Int -> HashMap k v -> LookupRes v
+lookupRecordCollision :: Eq k => Hash -> k -> Int -> HashMap k v -> LookupRes v
 lookupRecordCollision !_ !_ !_ Empty = Absent
 lookupRecordCollision h k _ (Leaf hx (L kx x))
     | h == hx && k == kx = Present x (-1)
@@ -501,7 +499,6 @@ lookupDefault def k t = case lookup k t of
     Just v -> v
     _      -> def
 {-# INLINABLE lookupDefault #-}
-
 
 -- | /O(log n)/ Return the value to which the specified key is mapped.
 -- Calls 'error' if this map contains no mapping for the key.
@@ -832,9 +829,9 @@ delete k0 m0 = go h0 k0 0 m0
     go h k s t@(Full ary) =
         let !st   = A.index ary i
             !st' = go h k (s+bitsPerSubkey) st
-         in if st' `ptrEq` st
-             then t
-             else case st' of
+        in if st' `ptrEq` st
+           then t
+           else case st' of
             Empty ->
                 let ary' = A.delete ary i
                     bm   = fullNodeMask .&. complement (1 `unsafeShiftL` i)
--- a/Data/HashMap/Base.hs
+++ b/Data/HashMap/Base.hs
@@ -574,8 +574,7 @@ insert k0 v0 m0 = go h0 k0 v0 0 m0
 -- We can skip:
 --  - the key equality check on a Leaf
 --  - check for its existence in the array for a hash collision
-insertNewKey :: Hashable k
-             => Hash -> k -> v -> Int -> HashMap k v -> HashMap k v
+insertNewKey :: Hash -> k -> v -> Int -> HashMap k v -> HashMap k v
 insertNewKey !h !k x !_ Empty = Leaf h (L k x)
 insertNewKey h k x s (Leaf hy l@(L ky y))
     | hy == h = collision h l (L k x)
@@ -624,9 +623,7 @@ insertNewKey h k x s t@(Collision hy v)
 --
 -- We can skip:
 --  - the key equality check on a Leaf, we know the leaf must be for this key
-insertKeyExists
-  :: Hashable k
-  => Int -> Hash -> k -> v -> Int -> HashMap k v -> HashMap k v
+insertKeyExists :: Int -> Hash -> k -> v -> Int -> HashMap k v -> HashMap k v
 insertKeyExists _collPos h k x s t@(Leaf hy (L ky y))
     | hy == h = if x `ptrEq` y
                 then t
@@ -864,7 +861,7 @@ delete k0 m0 = go h0 k0 0 m0
 --
 -- We can skip:
 --  - the key equality check on the leaf, if we reach a leaf it must be the key
-deleteKeyExists :: Hashable k => Int -> Hash -> k -> Int -> HashMap k v -> HashMap k v
+deleteKeyExists :: Int -> Hash -> k -> Int -> HashMap k v -> HashMap k v
 deleteKeyExists _collPos h _ _ t@(Leaf hy (L _ _))
   -- TODO(mrenaud): Can this comparison be removed?
     | hy == h = Empty
--- a/Data/HashMap/Base.hs
+++ b/Data/HashMap/Base.hs
@@ -830,8 +830,8 @@ delete k0 m0 = go h0 k0 0 m0
         let !st   = A.index ary i
             !st' = go h k (s+bitsPerSubkey) st
         in if st' `ptrEq` st
-           then t
-           else case st' of
+            then t
+            else case st' of
             Empty ->
                 let ary' = A.delete ary i
                     bm   = fullNodeMask .&. complement (1 `unsafeShiftL` i)
--- a/Data/HashMap/Base.hs
+++ b/Data/HashMap/Base.hs
@@ -25,9 +25,11 @@ module Data.HashMap.Base
     , insertWith
     , unsafeInsert
     , delete
+    , delete'
     , adjust
     , update
     , alter
+    , alterF
 
       -- * Combine
       -- ** Union
@@ -89,6 +91,15 @@ module Data.HashMap.Base
     , updateOrConcatWithKey
     , filterMapAux
     , equalKeys
+    , lookup'
+    , lookupRecordCollision
+    , insert'
+    , insertNewKey
+    , insertExistingKeyNoCollision
+    , insertExistingKeyCollision
+    , deleteNoCollision
+    , deleteCollision
+    , ptrEq
     ) where
 
 #if __GLASGOW_HASKELL__ < 710
@@ -431,22 +442,59 @@ member k m = case lookup k m of
 -- | /O(log n)/ Return the value to which the specified key is mapped,
 -- or 'Nothing' if this map contains no mapping for the key.
 lookup :: (Eq k, Hashable k) => k -> HashMap k v -> Maybe v
-lookup k0 m0 = go h0 k0 0 m0
-  where
-    h0 = hash k0
-    go !_ !_ !_ Empty = Nothing
-    go h k _ (Leaf hx (L kx x))
-        | h == hx && k == kx = Just x  -- TODO: Split test in two
-        | otherwise          = Nothing
-    go h k s (BitmapIndexed b v)
-        | b .&. m == 0 = Nothing
-        | otherwise    = go h k (s+bitsPerSubkey) (A.index v (sparseIndex b m))
-      where m = mask h s
-    go h k s (Full v) = go h k (s+bitsPerSubkey) (A.index v (index h s))
-    go h k _ (Collision hx v)
-        | h == hx   = lookupInArray k v
-        | otherwise = Nothing
-{-# INLINABLE lookup #-}
+lookup k m = lookup' (hash k) k 0 m
+{-# INLINE lookup #-}
+
+-- Internal helper for lookup. This version takes the precomputed hash so
+-- that functions that make multiple calls to lookup and related functions
+-- (insert, delete) only need to calculate the hash once.
+lookup' :: (Eq k, Hashable k) => Hash -> k -> Int -> HashMap k v -> Maybe v
+lookup' !_ !_ !_ Empty = Nothing
+lookup' h k _ (Leaf hx (L kx x))
+    | h == hx && k == kx = Just x  -- TODO: Split test in two
+    | otherwise          = Nothing
+lookup' h k s (BitmapIndexed b v)
+    | b .&. m == 0 = Nothing
+    | otherwise    = lookup' h k (s+bitsPerSubkey) (A.index v (sparseIndex b m))
+  where m = mask h s
+lookup' h k s (Full v) = lookup' h k (s+bitsPerSubkey) (A.index v (index h s))
+lookup' h k _ (Collision hx v)
+    | h == hx   = lookupInArray k v
+    | otherwise = Nothing
+{-# INLINABLE lookup' #-}
+
+-- Internal helper for lookup. This version takes the precomputed hash so
+-- that functions that make multiple calls to lookup and related functions
+-- (insert, delete) only need to calculate the hash once.
+--
+-- It is used by 'alterF' so that hash computation and key comparison only needs
+-- to be performed once. With this information you can use the more optimized
+-- versions of insert ('insertNewKey', 'insertExistingKeyNoCollision',
+-- 'insertExistingKeyCollision') and delete ('deleteNoCollision',
+-- 'deleteCollision').
+--
+-- Outcomes:
+--   Key not in map           => Nothing
+--   Key in map, no collision => Just (v, Nothing)
+--   Key in map, collision    => Just (v, position)
+lookupRecordCollision
+  :: (Eq k, Hashable k)
+  => Hash -> k -> Int -> HashMap k v -> Maybe (v, Maybe Int)
+lookupRecordCollision !_ !_ !_ Empty = Nothing
+lookupRecordCollision h k _ (Leaf hx (L kx x))
+    | h == hx && k == kx = Just (x, Nothing)
+    | otherwise          = Nothing
+lookupRecordCollision h k s (BitmapIndexed b v)
+    | b .&. m == 0 = Nothing
+    | otherwise    =
+        lookupRecordCollision h k (s+bitsPerSubkey) (A.index v (sparseIndex b m))
+  where m = mask h s
+lookupRecordCollision h k s (Full v) =
+  lookupRecordCollision h k (s+bitsPerSubkey) (A.index v (index h s))
+lookupRecordCollision h k _ (Collision hx v)
+    | h == hx   = fmap (fmap Just) $ lookupInArrayWithPosition k v
+    | otherwise = Nothing
+{-# INLINABLE lookupRecordCollision #-}
 
 -- | /O(log n)/ Return the value to which the specified key is mapped,
 -- or the default value if this map contains no mapping for the key.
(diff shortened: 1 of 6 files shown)


Data/HashMap/Base.hs
```diff
@@ -488,40 +536,184 @@ bitmapIndexedOrFull b ary
...
+{-# INLINABLE insert' #-}
+
+-- Insert optimized for the case when we know the key is not in the map.
+--
+-- It is only valid to call this when the key does not exist in the map.
+--
+-- We can skip:
+--  - the key equality check on a Leaf
+--  - check for its existance in the array for a hash collision
+insertNewKey :: Hashable k
```
treeowl: Spelled existence.

Data/HashMap/Base.hs
```diff
@@ -488,40 +536,184 @@ bitmapIndexedOrFull b ary
...
+-- Insert optimized for the case when we know the key is in the map and does not
+-- have a hash collision.
+--
+-- It is only valid to call this when the key exists in the map and does not
+-- have a hash collision with another key. This information can be obtained
+-- from 'lookupRecordCollision'.
+--
+-- We can skip:
+--  - the key equality check on a Leaf, we know the leaf must be for this key.
+insertExistingKeyNoCollision
```
treeowl: I think you can merge this with the version for collisions. Just use a negative index when there's not supposed to be a collision; you can test the sign when you get to the bottom for sanity checking.

Data/HashMap/Base.hs
```diff
@@ -658,57 +850,175 @@ unsafeInsertWith f k0 v0 m0 = runST (go h0 k0 v0 0 m0)
...
+-- Delete optimized for the case when we know the key is in the map and there is
+-- no collision.
+--
+-- It is only valid to call this when the key exists in the map and does not
+-- have a hash collision with another key. This information can be obtained
+-- from 'lookupRecordCollision'.
+--
+-- We can skip:
+--  - the key equality check on the leaf, if we reach a leaf it must be the key
+deleteNoCollision :: Hashable k => Hash -> k -> Int -> HashMap k v -> HashMap k v
```
treeowl: Again, I think this can be merged with the collision case.

Data/HashMap/Base.hs
```diff
@@ -1209,6 +1575,19 @@ lookupInArray k0 ary0 = go k0 ary0 0 (A.length ary0)
                 | otherwise -> go k ary (i+1) n
 {-# INLINABLE lookupInArray #-}
 
+-- | /O(n)/ Lookup the value associated with the given key in this
+-- array.  Returns 'Nothing' if the key wasn't found.
+lookupInArrayWithPosition :: Eq k => k -> A.Array (Leaf k v) -> Maybe (v, Int)
```
treeowl: What is this for?

Data/HashMap/Base.hs
```diff
@@ -658,57 +850,175 @@ unsafeInsertWith f k0 v0 m0 = runST (go h0 k0 v0 0 m0)
...
+-- Delete optimized for the case when we know the key is in the map and there is
+-- a hash collision at position @collPos@.
+--
+-- It is only valid to call this when the key exists in the map and has a hash
+-- collision with another key. This information can be obtained from
+-- 'lookupRecordCollision'.
+--
+-- We can skip:
+--  - the key equality check on the leaf, if we reach a leaf it must be the key
+deleteCollision :: (Eq k, Hashable k) => Int -> Hash -> k -> Int -> HashMap k v -> HashMap k v
```
treeowl: Why is there a constraint on this type?

Data/HashMap/Base.hs
treeowl: Also, why does this type have a constraint?

Data/HashMap/Base.hs
treeowl: And why is there a constraint?

Data/HashMap/Base.hs
m-renaud: Done.

Data/HashMap/Base.hs
m-renaud: Its used to lookup the key in the collision array (same as `lookupInArray`) but returns the value along with the collision position. It should be more clear now that I've switched to the `LookupRes a` type. It's called from the `Collision` case in `lookupRecordCollision`.

Data/HashMap/Base.hs
```diff
@@ -431,22 +442,59 @@ member k m = case lookup k m of
...
+-- 'insertExistingKeyCollision') and delete ('deleteNoCollision',
+-- 'deleteCollision').
+--
+-- Outcomes:
+--   Key not in map           => Nothing
+--   Key in map, no collision => Just (v, Nothing)
+--   Key in map, collision    => Just (v, position)
+lookupRecordCollision
+  :: (Eq k, Hashable k)
+  => Hash -> k -> Int -> HashMap k v -> Maybe (v, Maybe Int)
```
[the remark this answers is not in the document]
m-renaud: Done.

Another option would be to combine the `Alone` and `Collided` case and just set `-1` here. So you'd have:
```haskell
data LookupRes a = Absent | Present a !Int
```

Where `noCollision v = Present v (-1)` and `withCollision v collPos = Present v collPos`. This would simplify the `alterF` implementation by not having to case match on `Alone` and `Collided` since the alone case we need to pass `-1` anyways. Instead you would have in the case of delete:

```haskell
case lookupRes of
  Absent -> m
  Present _ collPos = deleteKeyExists collPos h k 0 m
```

What do you think?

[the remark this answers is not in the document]
treeowl: Sorry, I guess that should be

```haskell
data LookupRes a = Absent | Alone a | Collided !Int a
```

Or even just

```haskell
data LookupRes a = Absent | Present !Int a
```

with negative values meaning no collision. That's probably better.

treeowl:
I had the same idea, so yes, do that.

m-renaud:
Removed `lookup'`, `insert'`, `delete'`, unnecessary `Hashable` constraint, and simplified code.

Data/HashMap/Base.hs
```diff
@@ -689,9 +832,9 @@ delete k0 m0 = go h0 k0 0 m0
     go h k s t@(Full ary) =
         let !st   = A.index ary i
             !st' = go h k (s+bitsPerSubkey) st
-        in if st' `ptrEq` st
-            then t
-            else case st' of
+         in if st' `ptrEq` st
```
treeowl: Fix formatting.

Data/HashMap/Base.hs
```diff
@@ -448,6 +455,43 @@ lookup k0 m0 = go h0 k0 0 m0
...
+-- versions of insert ('insertNewKey', 'insertKeyExists') and delete
+-- ('deleteKeyExists')
+--
+-- Outcomes:
+--   Key not in map           => Absent
+--   Key in map, no collision => Alone v
+--   Key in map, collision    => Collide v position
+lookupRecordCollision
+  :: (Eq k, Hashable k)
+  => Hash -> k -> Int -> HashMap k v -> LookupRes v
```
treeowl: What is the `Int` argument? Why a `Hashable` constraint?

Data/HashMap/Base.hs
```diff
@@ -89,6 +90,12 @@ module Data.HashMap.Base
     , updateOrConcatWithKey
     , filterMapAux
     , equalKeys
+    , lookupRecordCollision
+    , LookupRes(..)
+    , insertNewKey
+    , insertKeyExists
+    , deleteKeyExists
+    , ptrEq
```
treeowl: Why are you adding these exports? I think we should do so in a separate PR if at all.

Data/HashMap/Base.hs
```diff
@@ -458,6 +502,7 @@ lookupDefault def k t = case lookup k t of
     _      -> def
 {-# INLINABLE lookupDefault #-}
 
+
```
treeowl: Kill the extra blank line.

Data/HashMap/Base.hs
```diff
@@ -751,12 +953,58 @@ update f = alter (>>= f)
 -- absence thereof. @alter@ can be used to insert, delete, or update a value in a
 -- map. In short : @'lookup' k ('alter' f k m) = f ('lookup' k m)@.
 alter :: (Eq k, Hashable k) => (Maybe v -> Maybe v) -> k -> HashMap k v -> HashMap k v
+-- TODO(m-renaud): Consider using specialized insert and delete for alter.
```
treeowl: This implementation of `alter` is ridiculous indeed. We will need some benchmarks with `Int` keys to make sure we don't get significant regressions by doing the sensible thing here.

treeowl:
Another possibility might be to make the lookup function take success and failure continuations. Then we can likely use the same underlying function to implement both `lookup` and `alterF`.

Data/HashMap/Strict.hs
```diff
@@ -249,6 +254,51 @@ alter f k m =
     Just v  -> insert k v m
 {-# INLINABLE alter #-}
 
+-- | /O(log n)/  The expression (@'alterF' f k map@) alters the value @x@ at
+-- @k@, or absence thereof. @alterF@ can be used to insert, delete, or update
+-- a value in a map.
+--
+-- @since 0.2.9
```
treeowl: Docs should mention that `alterF` is a flipped version of `Control.Lens.At.at`.

Data/HashMap/Base.hs
m-renaud: They're needed for the `alterF` version in the `Strict` module that forces the new value before its inserted into the map.

Data/HashMap/Base.hs
m-renaud: Done.

Data/HashMap/Base.hs
m-renaud: Done.

Data/HashMap/Base.hs
m-renaud: Int argument is needed for the BitMapIndexed and Full case. Hashable constraint removed.

Data/HashMap/Base.hs
m-renaud: I'll fix alter as a follow-up to this PR.

Data/HashMap/Base.hs
```diff
@@ -690,8 +830,8 @@ delete k0 m0 = go h0 k0 0 m0
         let !st   = A.index ary i
             !st' = go h k (s+bitsPerSubkey) st
         in if st' `ptrEq` st
-            then t
-            else case st' of
+           then t
```
m-renaud: I thought I fixed this, apparently not.

Data/HashMap/Base.hs
treeowl: Ah, of course. Silly me.

Data/HashMap/Base.hs
```diff
@@ -523,6 +565,104 @@ insert k0 v0 m0 = go h0 k0 v0 0 m0
...
+      mary <- A.new_ (n + 1)
+      A.copy ary 0 mary 0 n
+      A.write mary n leaf
+      return mary
+{-# INLINABLE insertNewKey #-}
+
+
+-- Insert optimized for the case when we know the key is in the map and does not
+-- have a hash collision.
+--
```
treeowl: This comment is out of date.

Data/HashMap/Base.hs
```diff
@@ -523,6 +565,104 @@ insert k0 v0 m0 = go h0 k0 v0 0 m0
...
+-- It is only valid to call this when the key exists in the map and you know the
+-- hash collision position if there was one. This information can be obtained
+-- from 'lookupRecordCollision'. If there is no collision pass (-1) as collPos
+-- (first argument).
+--
+-- We can skip:
+--  - the key equality check on a Leaf, we know the leaf must be for this key
```
treeowl: This bit of comment is poorly integrated.

Data/HashMap/Base.hs
treeowl: It's still changed.

Data/HashMap/Base.hs
m-renaud: Updated.

Data/HashMap/Base.hs
m-renaud: Changed it to a full sentence.

Data/HashMap/Base.hs
m-renaud: Gah! Hopefully fixed now.

Data/HashMap/Strict.hs
m-renaud: Done (in Base and Strict modules).

Data/HashMap/Base.hs
```diff
@@ -751,12 +949,61 @@ update f = alter (>>= f)
...
+    ------------------------------
+    -- Update value
+    Just v' -> case lookupRes of
+
+      -- Key did not exist before, insert v' under a new key
+      Absent -> insertNewKey h k v' 0 m
+
+      -- Key existed before, no hash collision
+      Present v collPos ->
+        -- TODO(m-renaud): Verify ptrEq is valid here.
```
m-renaud: @treeowl is this valid? aka are `v` and `v'` in whnf here? In theory I'm familiar with what that means: since the `Present` constructor is destructured then it should also force `v` to whnf, and pattern matching on `Just v'` above should do the same. Is that correct?

Data/HashMap/Strict.hs
```diff
@@ -249,6 +254,54 @@ alter f k m =
...
+      Absent -> insertNewKey h k v' 0 m
+
+      -- Key existed before, no hash collision
+      Present v collPos ->
+        -- TODO(m-renaud): Verify ptrEq is valid here.
+        if v `ptrEq` v'
+        -- If the value is identical, no-op
+        then m
+        -- If the value changed, update the value.
+        else v' `seq` insertKeyExists collPos h k v' 0 m
```
m-renaud: @treeowl Is the use of `seq` here fine or should I just use bang patterns on the `Just` pattern match above. My thought was it may be more efficient to wait until after the `ptrEq` check but it may not matter. WDYT?

Data/HashMap/Base.hs
treeowl: The pointer equality test is valid even if they're not in WHNF--it's just less reliable. Remember the rule: yes means yes; no means no information.

m-renaud: Gotcha, thanks for clarifying.

treeowl: Take the to-do comments about pointer equality out?

Data/HashMap/Strict.hs
```diff
@@ -249,6 +254,54 @@ alter f k m =
     Just v  -> insert k v m
 {-# INLINABLE alter #-}
 
+-- | /O(log n)/  The expression (@'alterF' f k map@) alters the value @x@ at
+-- @k@, or absence thereof. @alterF@ can be used to insert, delete, or update
+-- a value in a map.
+--
+-- Note: 'alterF' is a flipped version of the 'at' combinator from
+-- 'Control.Lens.At'.
+--
```
treeowl: I think you need a URL; we don't depend on `lens` here!

Data/HashMap/Base.hs
```diff
@@ -523,6 +565,103 @@ insert k0 v0 m0 = go h0 k0 v0 0 m0
...
+--
+-- It is only valid to call this when the key exists in the map and you know the
+-- hash collision position if there was one. This information can be obtained
+-- from 'lookupRecordCollision'. If there is no collision pass (-1) as collPos
+-- (first argument).
+--
+-- We can skip the key equality check on a Leaf because we know the leaf must be
+-- for this key.
+insertKeyExists :: Int -> Hash -> k -> v -> Int -> HashMap k v -> HashMap k v
+insertKeyExists _collPos h k x s t@(Leaf hy (L ky y))
```
treeowl: Please document what the second `Int` argument means. That's not obvious.

Data/HashMap/Strict.hs
m-renaud: Fixed. This is also broken in `containers`.

Data/HashMap/Base.hs
m-renaud: Removed.

Data/HashMap/Base.hs
```diff
@@ -448,6 +455,41 @@ lookup k0 m0 = go h0 k0 0 m0
...
+-- It is used by 'alterF' so that hash computation and key comparison only needs
+-- to be performed once. With this information you can use the more optimized
+-- versions of insert ('insertNewKey', 'insertKeyExists') and delete
+-- ('deleteKeyExists')
+--
+-- Outcomes:
+--   Key not in map           => Absent
+--   Key in map, no collision => Alone v
+--   Key in map, collision    => Collide v position
+lookupRecordCollision :: Eq k => Hash -> k -> Int -> HashMap k v -> LookupRes v
```
treeowl: What is the `Int` argument? That's not obvious!

treeowl:
Oh, I see. Those extra Int parameters are because you didn't split off go
functions like the usual things. I think you should. Keep things modular.

treeowl:
I would suggest that you write a function

{-# INLINE lookupCont #-}
lookupCont :: Eq k => r -> (Int -> v -> r) -> Hash -> k -> HashMap k v -> r

Now

lookup k m = lookupCont Nothing (\_ v -> Just v) (hash k) m
lookupWhatever k m = lookupCont Absent Present (hash k) m

This obviously cuts down the source code size.

treeowl:
Your benchmark results don't really look so great compared to `alter`, even when the keys are *strings*. We need to understand why.

m-renaud:
> Your benchmark results don't really look so great compared to alter

I think this could be explained by the use of the `Just` functor in the `alterF` benchmarks (no idea why I didn't use `Identity` to begin with). This incurs an extra constructor call and pattern match (`fromJust`) on each insertion/deletion/etc. I imagine that using a function which is simply a newtype wrapper (`Identity` for example) the difference will go away. I'll update the benchmarks and re-run to confirm.

treeowl:
Using the right functor helps, but performance is still cruddy. I marked a
bunch of arguments strict, rearranged pointer equality tests, and removed
several dead branches. Oddly, I've gotten insertion faster than alter for
ByteString, but not for String. I'll have another look later.

m-renaud:
So, I replaced the implementation with the naive implementation (same as `alter`) and the benchmark results improved. They're still not as good as `alter` but they're better than all of custom code that avoids computing the hash multiple times and uses optimized inserts and deletes given info about the key and collisions. Weird.

**alter**
```
benchmarking HashMap/alterInsert/String
time                 1.544 ms   (1.519 ms .. 1.571 ms)
                     0.997 R²   (0.996 R² .. 0.999 R²)
mean                 1.403 ms   (1.371 ms .. 1.427 ms)
std dev              84.11 μs   (58.81 μs .. 140.3 μs)
variance introduced by outliers: 45% (moderately inflated)

benchmarking HashMap/alterInsert/ByteString
time                 1.579 ms   (1.568 ms .. 1.593 ms)
                     0.999 R²   (0.998 R² .. 0.999 R²)
mean                 1.441 ms   (1.420 ms .. 1.462 ms)
std dev              69.51 μs   (59.86 μs .. 80.24 μs)
variance introduced by outliers: 34% (moderately inflated)

benchmarking HashMap/alterInsert/Int
time                 1.088 ms   (1.064 ms .. 1.121 ms)
                     0.997 R²   (0.995 R² .. 0.999 R²)
mean                 987.5 μs   (972.8 μs .. 1.004 ms)
std dev              46.62 μs   (35.65 μs .. 67.75 μs)
variance introduced by outliers: 35% (moderately inflated)

benchmarking HashMap/alterInsert-dup/String
time                 771.1 μs   (763.7 μs .. 777.8 μs)
                     0.999 R²   (0.998 R² .. 0.999 R²)
mean                 718.7 μs   (709.1 μs .. 728.3 μs)
std dev              29.82 μs   (25.41 μs .. 35.75 μs)
variance introduced by outliers: 31% (moderately inflated)

benchmarking HashMap/alterInsert-dup/ByteString
time                 319.5 μs   (317.6 μs .. 322.0 μs)
                     0.998 R²   (0.996 R² .. 1.000 R²)
mean                 296.6 μs   (290.7 μs .. 303.0 μs)
std dev              18.68 μs   (14.57 μs .. 24.97 μs)
variance introduced by outliers: 56% (severely inflated)

benchmarking HashMap/alterInsert-dup/Int
time                 588.9 μs   (581.3 μs .. 593.3 μs)
                     0.998 R²   (0.997 R² .. 0.999 R²)
mean                 521.2 μs   (503.2 μs .. 536.3 μs)
std dev              47.30 μs   (36.95 μs .. 55.63 μs)
variance introduced by outliers: 70% (severely inflated)

benchmarking HashMap/alterDelete/String
time                 1.252 ms   (1.246 ms .. 1.258 ms)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 1.181 ms   (1.168 ms .. 1.193 ms)
std dev              39.06 μs   (31.66 μs .. 49.15 μs)
variance introduced by outliers: 20% (moderately inflated)

benchmarking HashMap/alterDelete/ByteString
time                 798.7 μs   (775.2 μs .. 844.2 μs)
                     0.973 R²   (0.931 R² .. 1.000 R²)
mean                 730.9 μs   (708.1 μs .. 794.2 μs)
std dev              104.1 μs   (41.34 μs .. 192.7 μs)
variance introduced by outliers: 85% (severely inflated)

benchmarking HashMap/alterDelete/Int
time                 537.2 μs   (535.7 μs .. 538.9 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 501.5 μs   (496.0 μs .. 506.8 μs)
std dev              17.10 μs   (14.70 μs .. 19.52 μs)
variance introduced by outliers: 26% (moderately inflated)

benchmarking HashMap/alterDelete-miss/String
time                 387.6 μs   (384.2 μs .. 390.1 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 352.6 μs   (347.0 μs .. 358.9 μs)
std dev              17.21 μs   (14.78 μs .. 20.35 μs)
variance introduced by outliers: 43% (moderately inflated)

benchmarking HashMap/alterDelete-miss/ByteString
time                 201.9 μs   (199.6 μs .. 203.6 μs)
                     0.999 R²   (0.997 R² .. 1.000 R²)
mean                 187.4 μs   (184.6 μs .. 190.0 μs)
std dev              7.937 μs   (6.466 μs .. 9.822 μs)
variance introduced by outliers: 39% (moderately inflated)

benchmarking HashMap/alterDelete-miss/Int
time                 323.3 μs   (314.0 μs .. 330.0 μs)
                     0.997 R²   (0.997 R² .. 0.999 R²)
mean                 288.8 μs   (282.8 μs .. 295.6 μs)
std dev              18.54 μs   (14.18 μs .. 23.29 μs)
variance introduced by outliers: 58% (severely inflated)
```

**alterF**

```
benchmarking HashMap/alterFInsert/String
time                 1.730 ms   (1.688 ms .. 1.761 ms)
                     0.996 R²   (0.993 R² .. 0.997 R²)
mean                 1.485 ms   (1.437 ms .. 1.529 ms)
std dev              131.4 μs   (112.0 μs .. 152.4 μs)
variance introduced by outliers: 63% (severely inflated)

benchmarking HashMap/alterFInsert/ByteString
time                 1.645 ms   (1.623 ms .. 1.668 ms)
                     0.998 R²   (0.997 R² .. 0.999 R²)
mean                 1.494 ms   (1.469 ms .. 1.517 ms)
std dev              78.77 μs   (67.09 μs .. 93.80 μs)
variance introduced by outliers: 38% (moderately inflated)

benchmarking HashMap/alterFInsert/Int
time                 1.211 ms   (1.176 ms .. 1.270 ms)
                     0.989 R²   (0.976 R² .. 0.997 R²)
mean                 1.163 ms   (1.141 ms .. 1.195 ms)
std dev              75.99 μs   (48.11 μs .. 128.2 μs)
variance introduced by outliers: 51% (severely inflated)

benchmarking HashMap/alterFInsert-dup/String
time                 755.5 μs   (752.8 μs .. 757.7 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 708.4 μs   (700.4 μs .. 715.0 μs)
std dev              23.95 μs   (21.26 μs .. 27.36 μs)
variance introduced by outliers: 24% (moderately inflated)

benchmarking HashMap/alterFInsert-dup/ByteString
time                 340.7 μs   (339.6 μs .. 341.9 μs)
                     1.000 R²   (0.999 R² .. 1.000 R²)
mean                 320.1 μs   (317.5 μs .. 322.6 μs)
std dev              7.662 μs   (6.720 μs .. 8.826 μs)
variance introduced by outliers: 15% (moderately inflated)

benchmarking HashMap/alterFInsert-dup/Int
time                 633.0 μs   (624.6 μs .. 639.7 μs)
                     0.998 R²   (0.997 R² .. 0.999 R²)
mean                 561.6 μs   (541.2 μs .. 576.5 μs)
std dev              49.41 μs   (37.66 μs .. 58.30 μs)
variance introduced by outliers: 70% (severely inflated)

benchmarking HashMap/alterFDelete/String
time                 1.289 ms   (1.275 ms .. 1.306 ms)
mean                 1.211 ms   (1.198 ms .. 1.224 ms)
std dev              41.74 μs   (32.50 μs .. 53.49 μs)
variance introduced by outliers: 22% (moderately inflated)

benchmarking HashMap/alterFDelete/ByteString
time                 817.6 μs   (813.6 μs .. 822.2 μs)
                     1.000 R²   (0.999 R² .. 1.000 R²)
mean                 768.1 μs   (760.3 μs .. 776.3 μs)
std dev              25.58 μs   (22.51 μs .. 29.96 μs)
variance introduced by outliers: 23% (moderately inflated)

benchmarking HashMap/alterFDelete/Int
time                 527.9 μs   (526.2 μs .. 529.2 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 485.3 μs   (477.9 μs .. 492.3 μs)
std dev              22.39 μs   (18.60 μs .. 26.73 μs)
variance introduced by outliers: 39% (moderately inflated)

benchmarking HashMap/alterFDelete-miss/String
time                 394.8 μs   (389.2 μs .. 401.3 μs)
mean                 360.2 μs   (354.4 μs .. 366.7 μs)
std dev              17.77 μs   (14.37 μs .. 23.38 μs)
variance introduced by outliers: 44% (moderately inflated)

benchmarking HashMap/alterFDelete-miss/ByteString
time                 187.6 μs   (182.8 μs .. 194.8 μs)
                     0.996 R²   (0.992 R² .. 1.000 R²)
mean                 169.4 μs   (166.8 μs .. 174.0 μs)
std dev              9.705 μs   (6.249 μs .. 16.97 μs)
variance introduced by outliers: 54% (severely inflated)

benchmarking HashMap/alterFDelete-miss/Int
time                 345.3 μs   (337.5 μs .. 354.7 μs)
                     0.997 R²   (0.996 R² .. 0.999 R²)
mean                 329.6 μs   (326.3 μs .. 333.2 μs)
std dev              10.66 μs   (9.120 μs .. 12.70 μs)
variance introduced by outliers: 25% (moderately inflated)
```

treeowl:
Hmmmm. Don't want the naive implementation. Want to figure out how to fix
this one. Might need some thinking.

treeowl:
Take a look at [this version](https://github.com/treeowl/unordered-containers/tree/m-renaud-alterF). It doesn't have the `go` functions you've added, but it seems to be significantly faster than yours. In particular, I've dropped a bunch of tests when we know what the results will be. Among other things, we never need pointer equality tests in the helper functions for `alterF` because `alterF` does all the pointer equality tests we need. I'm still not entirely satisfied with the performance, but it's a start.

m-renaud:
Thanks for taking a closer look at this! So, I integrated your changes (remove `go` functions, unnecessary `ptrEq` checks, and a few other unnecessary checks), but didn't really see much speedup in my benchmarks (see https://github.com/tibbe/unordered-containers/pull/180/commits/db0581d74dd249d44cd2b3acc4b7e438be67462b). I had originally written it without the `go` functions and didn't see a regression when switching to them. (Also, I don't _think_ I made a mistake merging but I'll have to take a closer look and could use another set of eyes).

I did try using unboxed sums for the `Absent/Present` case which seemed to help, but that's only available in the most recent GHC version. I haven't had time to put a commit together for that change yet either, its a somewhat intrusive change since `lookupCont` can't be used due to mixing lifted and unlifted types (or that's the error I was getting anyways).

And sorry for the delay, I've been out of town snowboarding.

treeowl:
Oh, I didn't remove go functions; I just was working on a version before
you added them. I really want those!

treeowl:
You can avoid code duplication using TypeInType. For GHC 8.2 and later,
make the underlying CPSed lookup function representation-polymorphic in the
result type. Note that the failure continuation will need type `Void# -> r`
or `(# #) -> r` in order to avoid a prohibited representation-polymorphic
argument. You can use the same function definition for all versions; only
the type signature changes.

m-renaud:
> Oh, I didn't remove go functions ...  I really want those!

Gotcha, I was a little confused, that's what I thought :)

> You can avoid code duplication using TypeInType

Cool, I'll look into that, I've never used TypeInType before and this is my first time working with unboxed sums.

treeowl:
Basic idea:

{-# language TypeInType #-}
import GHC.Exts ( TYPE )

lookupCont :: forall rep (r :: TYPE rep) k v. Eq k => ((# #) -> r) -> (v ->
Int -> r) -> Hash -> k -> HashMap k v -> r

Now the result can be Maybe a, or LookupRes a, or (# (# #) | a #), or (# (#
#) | (# Int, a #) #), or whatever.

treeowl:
I *finally* understand a big part of the balancing act: do we or don't we look up eagerly? The definition of `alter` does *not*. If the passed function is *lazy* (doesn't care whether the key is already in the map), then it simply *skips* the lookup, which is super-cheap. The version of `alterF` with unboxed sums *does*, I'm pretty sure. This is lousy when `f` is completely lazy, but better when it's not. Blargh!

treeowl:
Side note: I believe the calls to `error` in the helper functions are causing some allocations as a result of the call stack mechanism. Also, I don't think the `go` functions are being lifted to the top level as they should be. One option is to replace the `error` calls with calls to `GHC.Err.errorWithoutCallStack` or whatever it's called, when GHC is recent enough to have that. And you can lift the `go` functions to the top level manually, which still confines the `0`-passing geographically. I'm not sure if the `go` bit matters or not.

treeowl:
A possible compromise: use the eager-lookup version in general, but use rewrite rules to catch certain special cases. I believe that practically speaking, we only have to care about this laziness for the `Identity` functor. We have

```haskell
alterF f k m = (<$> f mv) $ \fres -> ...
  where ...
```

Let's break that up a bit:

```haskell
process :: (Hashable k, Eq k) => k -> Identity (Maybe a) -> HashMap k a -> (Maybe a -> Maybe a) -> Identity (HashMap k a)
process fmv _ f = f <$> fmv
{#- INLINE [1] process #-}

alterF f k m = process (f mv) m $ \fres -> ...
  where ...
```

Now we can add rewrite rules:

```haskell
{-# RULES
"alterFinsert" forall k v m. process k (Identity (Just v)) m _ = Identity (insert k v m)
"alterFdelete" forall k m. process k (Identity Nothing) m _ = Identity (delete k m)
  #-}
```

m-renaud:
> I believe the calls to error in the helper functions are causing some allocations

I can look into using the non-call stack version as you suggested.

> do we or don't we look up eagerly? The definition of alter does not.

That's a great point, I hadn't considered that!

>  If the passed function is lazy (doesn't care whether the key is already in the map)

I'm curious how often this actually happens in practice. If you don't care whether the key is in the map or not (which also means you don't care what the value is), then all you have is a simple `insert` or `delete`. This means you can just perform the insert or delete, and then if you realllllly need the map in the functor then just put the value into the functor afterwards:

```haskell
foo :: Functor f => Maybe v -> f (Maybe v)
bar :: Functor f => f v
baz :: Functor f => f ()


alterF (\_ -> foo (Just v) k myMap == flip (M.insert k) myMap <$> bar
alterF (\_ -> foo Nothing) k myMap == const (M.delete k myMap) <$> baz
```

If the functions existed, they would be `insertF` and `deleteF`:

```haskell
insertF :: (Functor f, Ord k) => k -> f v -> HashMap k v -> f (HashMap k v)
insertF k fv m = flip (M.insert k) m <$> fv

deleteF :: (Functor f, Ord k) => k -> f () -> HashMap k v -> f (HashMap k v)
deleteF k fv m = const (M.delete k m) <$> fv
```

> A possible compromise:... we only have to care about this laziness for the Identity functor

That's reasonable, although since its so narrowly scoped I'd like to push for doing it in a follow up.

treeowl:
Nobody sensible is going to use `const ...` in an explicit call to
`alterF`. But we want `Control.Lens.At` to use `alterF` in the
implementation of `At (HashMap k a)`. The lens users will then end up using
it for all sorts of insertions and deletions, unfortunately. Indeed, they
probably are doing so already!

treeowl:
Making the RULES work is really fussy, by the way. It turns out you can't
successfully use the Identity newtype constructor on the LHS of a rule. You
have to use `coerce` instead, specialized to the appropriate type. The
RULES approach only works for specific functors we select, which is a bit
sad, but I do suspect that to satisfy the lensers we'll only really need to
make sure Identity is good for inserting and deleting.

m-renaud:
> Nobody sensible is going to use `const ...` in an explicit call to `alterF`

Haha, yeah, I didn't expect so :P I just wanted to illustrate that if you _really_ needed the "effects" of the functor (either after the deletion or to determine the value to update) that you can do it without alterF.

> The lens users will then end up using it for all sorts of insertions and deletions, unfortunately. Indeed, they probably are doing so already!

You're probably right. In the case of the lens usage would the `Identity` re-write rules actually help? Or would a different functor be used a lot of the time?

I'm not opposed to refactoring the code and adding RULES, but if this is about the best performance we can get for non-specific cases then I'd prefer to check this in and build off of it in another PR.

Also, should we go back to using non-unboxed sums for `GLASGOW_HASKELL < 802`? The performance isn't as great due to the additional `LookupRes` constructors but I don't think there's anything we can do about that :/ WDYT?

treeowl:
I have a couple ideas I'd like to try out. lens makes heavy use of both
Identity and Const; nothing tricky there.

treeowl:
Yeah, I'm pretty sure I can make the horrible `RULES` hack work, and with a bit less mess than the way I suggested above:

Write a function that looks up eagerly (unboxing the lookup result). Call that `alterFEager` or something. Write a lazy version called `alterF`. Now write

```haskell
{-# NOINLINE alterF #-}
{-# RULES
  "weirdify" forall f k m.  alterF f k m = alterFWeird (f (error "Internal error")) f k m
 #-}

-- This is a very unsafe sort of alterF. It must only be called with a first argument equal
-- to the result of applying its second argument to bottom.
alterFWeird :: (Eq k, Hashable k)
  => f (Maybe v) -> (Maybe v -> f (Maybe v)) -> k -> HashMap k m -> f (HashMap k m)
alterFWeird _ f k m = alterFEager f k m
{-# INLINE [1] alterFWeird #-}
{-# RULES
  "alterFdelete" forall (f :: Maybe v -> Identity (Maybe v)) k m.
      alterFWeird (coerce Nothing) _ k m = Identity (delete k m)
  "alterFinsert" forall (f :: Maybe v -> Identity (Maybe v)) k m x.
      alterFWeird (coerce (Just x)) _ k m = Identity (insert k x m)
  #-}
```

When someone using `Control.Lens.At simply` inserts or deletes in `Identity` (by far the most likely functor for that), `alterFWeird` will be rewritten to a simple insertion or deletion. If they insert or delete in some other functor, they'll only pay the extra cost of a single unnecessary lookup, which isn't really terrible. If they do anything else, `alterFWeird` will inline, eliminating the unnecessary `f bottom` thunk, and turn into your highly-optimized eager version. Finally, if someone is compiling without optimizations, they'll get the so-so lazy version which will be decent for insertions and deletions and won't be totally terrible for other operations.

m-renaud:
Interesting, I'll have to take a closer look when I have a larger block of
free time. I like the idea though and the behavior matches what I would
expect.

treeowl:
Actually, we can probably go even further by adding more arguments to `alterFWeird`. In particular, suppose we have

```haskell
hiddenBottom :: a
hiddenBottom = error "whoops"
{-# NOINLINE hiddenBottom #-}
```
Now we can call `alterFWeird (f undefined) (f Nothing) (f (Just hiddenBottom)) f k m`. Then we can match in `RULES`:

```haskell
-- approximately
alterFWeird _whatever (Identity Nothing) (Identity (Just (g hiddenBottom))) f k m = adjust g k m
```

treeowl:
Subsumed by #181.

PR #181, opened 2018-01-31
Optimized alterF

treeowl:


treeowl:
@m-renaud I realized the precise scheme I described in a comment on your PR won't work quite right, but I *think* this will. The benchmarks look pretty reasonable to me. What do you think?

m-renaud:
I don't see the benchmarks in the commit description, could you add them either to the commit or to the PR thread? I'm still grokking the changes to make sure everything looks reasonable. This is definitely at the edge of my current Haskelling ability though :)

m-renaud:
Also, TIL about https://wiki.haskell.org/Worker_wrapper, neat.

Data/HashMap/Base.hs
```diff
@@ -461,12 +461,12 @@ lookup :: (Eq k, Hashable k) => k -> HashMap k v -> Maybe v
 -- unboxed sums automatically. That seems likely to happen at some
 -- point (possibly as early as GHC 8.6) but for now we do it manually.
 lookup k m = case lookup# k m of
-  (# (# #) | #) -> Nothing
+  (# _ | #) -> Nothing
```
m-renaud: Does this have any impact on performance? I would hope not, but if it does would it be worth it to condition this on GHC versions that support nullary tuples? It would clutter the code quite a bit more though :/

Data/HashMap/Base.hs
```diff
@@ -751,12 +1102,129 @@ update f = alter (>>= f)
...
+alterF f !k !m = (<$> f mv) $ \fres ->
+  case fres of
+    Nothing -> delete' h k m
+    Just v' -> insert' h k v' m
+
+  where !h = hash k
+        mv = lookup' h k m
+{-# INLINABLE [0] alterF #-}
+
+bogus :: a
```
m-renaud: Can you add comments here and for bogus# describing why they're needed/how they're used?

treeowl:
I don't really trust my benchmarks. I'm hoping maybe you can run some. Speaking of benchmark trust, I believe we're using the Criterion "environments" wrong and consequently getting some very strong order-dependence in our benchmarks. That's no good! Specifically, it seems that the second test run on the same data will generally be faster.

Data/HashMap/Base.hs
```diff
@@ -431,22 +456,116 @@ member k m = case lookup k m of
...
+-- unboxed sums automatically. That seems likely to happen at some
+-- point (possibly as early as GHC 8.6) but for now we do it manually.
+lookup' h k m = case lookup'# h k m of
+  (# (# #) | #) -> Nothing
+  (# | a #) -> Just a
+{-# INLINE lookup' #-}
+
+lookup'# :: Eq k => Hash -> k -> HashMap k v -> (# (# #) | v #)
+lookup'# h k m = lookupCont (\_ -> (# (# #) | #)) (\v _i -> (# | v #)) h k m
+{-# INLINABLE lookup'# #-}
```
m-renaud: For my own edification, why are some things INLINE and others INLINABLE? Is there any rule to follow when deciding which one?

tests/HashMapProperties.hs
```diff
@@ -168,6 +169,25 @@ pAlterInsert k = M.alter (const $ Just 3) k `eq_` HM.alter (const $ Just 3) k
...
+pAlterFAdjust k =
+  fromJust . M.alterF (Just . fmap succ) k `eq_`
+  fromJust . HM.alterF (Just . fmap succ) k
+
+pAlterFInsert :: Key -> [(Key, Int)] -> Bool
+pAlterFInsert k =
+  fromJust . M.alterF (const . Just . Just $ 3) k `eq_`
+  fromJust . HM.alterF (const . Just . Just $ 3) k
+
+pAlterFDelete :: Key -> [(Key, Int)] -> Bool
```
m-renaud: Can you add benchmarks to show the difference between the rewritten `Identity` and a non rewritten version?

Data/HashMap/Base.hs
```diff
@@ -708,36 +988,107 @@ delete k0 m0 = go h0 k0 0 m0
...
+-- hash collision position if there was one. This information can be obtained
+-- from 'lookupRecordCollision'. If there is no collision pass (-1) as collPos.
+--
+-- We can skip:
+--  - the key equality check on the leaf, if we reach a leaf it must be the key
+deleteKeyExists :: Int -> Hash -> k -> HashMap k v -> HashMap k v
+deleteKeyExists !collPos0 !h0 !k0 !m0 = dkego collPos0 h0 k0 0 m0
+{-# NOINLINE deleteKeyExists #-}
+
+dkego :: Int -> Hash -> k -> Int -> HashMap k v -> HashMap k v
```
m-renaud: Any reason for making this top level instead of a in the `where` clause of deleteKeyExists?

Data/HashMap/Base.hs
```diff
@@ -619,6 +829,73 @@ insertWith f k0 v0 m0 = go h0 k0 v0 0 m0
...
+    go h k s t@(Collision hy v)
+        | h == hy   =
+            let !v' = insertModifyingArr x f k v
+            in if A.unsafeSameArray v v'
+               then t
+               else Collision h v'
+        | otherwise = go h k s $ BitmapIndexed (mask hy s) (A.singleton t)
+{-# INLINABLE insertModifying #-}
+
+insertModifyingArr :: Eq k => v -> (v -> (# v #)) -> k -> A.Array (Leaf k v)
```
m-renaud: Can you add a function comment here?

m-renaud:
> Speaking of benchmark trust...

I've been running benchmarks that had a lot of outliers on their own (multiple times) using criterion's bench group filters to get more stable numbers. It's somewhat time consuming but seemed to give more accurate results.

Data/HashMap/Base.hs
treeowl: I'm not sure. I have some questions about whether certain `letrec`s end up allocating closures. I'll probably have to benchmark or something.

tests/HashMapProperties.hs
treeowl: Not a bad idea. Need tests too, ideally.

Data/HashMap/Base.hs
treeowl: I'm not the best at that. `lookup'#` is in the nature of a "worker". Once `lookupCont` inlines into it, it will actually be fairly large. We therefore generally *don't* want to actually inline it at call sites. But we *do* want to make sure that GHC gives it an unfolding in the interface (`.hi`) file. That will allow the specializer to specialize it (in other modules) to the key types it's actually used with.

We follow GHC's own practice (when it performs the worker-wrapper transformation on *products*) of marking the wrappers `INLINE`. We really really want them to inline in most cases to prevent "boxing" (in this case the allocation of `Just` constructors). In *most* cases, inlining the wrapper will very quickly expose opportunities for simplification, and very likely *reduce* code size.

Data/HashMap/Base.hs
treeowl: I can try.

Data/HashMap/Base.hs
treeowl: It turns out they all support nullary tuples; I just got confused.

Data/HashMap/Base.hs
treeowl: Yes, good idea. I should also change the names. They are really just bottom values, and not really related to each other. `bogus#` is the name of an internal error with a strange type. `bogus` is used in a much more interesting way. Applying a function to a partially defined value lets us learn things about it. If GHC sees `f (Just bogus)`, where

```haskell
f (Just x) = Just E  -- E is some expression
```

then it will reduce that to `Just (E [x -> bogus])`, where the bit in brackets is a substitution. We can match on that `Just ...` result in `RULES`, learning things about `f` that let us rewrite the `alterF` form to something better.

Data/HashMap/Base.hs
```diff
@@ -751,12 +1083,131 @@ update f = alter (>>= f)
...
+-- version; otherwise someone could tell the difference using a lazy
+-- @f@ and a functor that is similar to Const but not actually Const.
+alterF f !k !m = (<$> f mv) $ \fres ->
+  case fres of
+    Nothing -> delete' h k m
+    Just v' -> insert' h k v' m
+
+  where !h = hash k
+        mv = lookup' h k m
+{-# INLINABLE [0] alterF #-}
```
treeowl: We should probably just say `NOINLINE` or `NOINLINE [0]`. This implementation should really only be used in code compiled without optimization.

Data/HashMap/Base.hs
```diff
@@ -751,12 +1083,131 @@ update f = alter (>>= f)
...
+
+#if MIN_VERSION_base(4,8,0)
+bogus :: a
+bogus = error "Data.HashMap.alterF internal error: hit bottom"
+
+bogus# :: (# #) -> (# a #)
+bogus# _ = error "Data.HashMap.alterF internal error: hit bottom#"
```
treeowl: Rename the boguses, and explain the function of each.

Data/HashMap/Base.hs
```diff
@@ -751,12 +1083,131 @@ update f = alter (>>= f)
...
+       => f (Maybe v)
+       -> f (Maybe v)
+       -> (Maybe v -> f (Maybe v)) -> k -> HashMap k v -> f (HashMap k v)
+alterFWeird _ _ f k m = alterFEager f k m
+{-# INLINE [0] alterFWeird #-}
+
+alterFEager :: (Functor f, Eq k, Hashable k)
+       => (Maybe v -> f (Maybe v)) -> k -> HashMap k v -> f (HashMap k v)
+alterFEager f !k m = (<$> f mv) $ \fres ->
+  case fres of
```
treeowl: Explain why we don't look up `k` in `m` on the outside, if indeed there's a good reason.

Data/HashMap/Base.hs
m-renaud: I see rules for `alterFWeird` but not for this, is the idea that rules should fire for `alterF` so the implementation here is rarely used?

Data/HashMap/Base.hs
m-renaud: On the outside meaning before this function is called and instead pass in `lookupRes`?

treeowl:
The first rule unconditionally rewrites alterF, so yes, that's the idea.

On Jan 31, 2018 9:35 AM, "Matt Renaud" <notifications@github.com> wrote:

*@m-renaud* commented on this pull request.
------------------------------

In Data/HashMap/Base.hs
<https://github.com/tibbe/unordered-containers/pull/181#discussion_r165069279>
:

> +-- @since 0.2.9
+alterF :: (Functor f, Eq k, Hashable k)
+       => (Maybe v -> f (Maybe v)) -> k -> HashMap k v -> f (HashMap k v)
+-- We only calculate the hash once, but unless this is rewritten
+-- by rules we may test for key equality multiple times.
+-- We force the value of the map for consistency with the rewritten
+-- version; otherwise someone could tell the difference using a lazy
+-- @f@ and a functor that is similar to Const but not actually Const.
+alterF f !k !m = (<$> f mv) $ \fres ->
+  case fres of
+    Nothing -> delete' h k m
+    Just v' -> insert' h k v' m
+
+  where !h = hash k
+        mv = lookup' h k m
+{-# INLINABLE [0] alterF #-}

I see rules for alterFWeird but not for this, is the idea that rules should
fire for alterF so the implementation here is rarely used?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/tibbe/unordered-containers/pull/181#discussion_r165069279>,
or mute the thread

Data/HashMap/Base.hs
treeowl: I mean the reason we don't have

```haskell
alterFEager f k m = case lookupRecordBlah k m of ...

File: Data/HashMap/Base.hs [hunk 10]
```diff
@@ -751,12 +1083,131 @@ update f = alter (>>= f)
...
+
+#if MIN_VERSION_base(4,8,0)
+bogus :: a
+bogus = error "Data.HashMap.alterF internal error: hit bottom"
+
+bogus# :: (# #) -> (# a #)
+bogus# _ = error "Data.HashMap.alterF internal error: hit bottom#"
+
+{-# RULES
+"alterFWeird" forall f k m. alterF f k m = alterFWeird (f Nothing) (f (Just bogus)) f k m
```
m-renaud: Oh here it is, I missed this when reading it the first time. Can you add a comment that we unconditionally rewrite alterF? Also consider dropping the rewrite part `alterF f k m = ...` to the next line so its separated visually from the forall part.

Data/HashMap/Strict.hs
```diff
@@ -249,6 +259,112 @@ alter f k m =
...
+  alterFWeird (coerce Nothing) (coerce (Just x)) f k m =
+    Identity (adjust (\a -> case runIdentity (f (Just a)) of
+                               Just a' -> a'
+                               Nothing -> bogus) k m)
+
+"alterFdelete" forall (f :: Maybe a -> Identity (Maybe a)) k m.
+  alterFWeird (coerce Nothing) (coerce Nothing) f k m = Identity (delete k m)
+
+"alterFlookup" forall _ign1 _ign2 (f :: Maybe a -> Const r (Maybe a)) k m.
+  alterFWeird _ign1 _ign2 f k m = k `seq` m `seq` Const (getConst (f (lookup k m)))
```
m-renaud: Out of curiosity, can you not use bang patterns in rewrite rules?

Data/HashMap/Base.hs
```diff
@@ -751,12 +1083,131 @@ update f = alter (>>= f)
...
+      -- Key existed before, no hash collision
+      Present v collPos ->
+        if v `ptrEq` v'
+        -- If the value is identical, no-op
+        then m
+        -- If the value changed, update the value.
+        else insertKeyExists collPos h k v' m
+
+  where !h = hash k
+        !lookupRes = lookupRecordCollision h k m
```
m-renaud: Any reason why this function doesn't have a version for `GLASGOW_HASKELL >= 802` that uses `lookupRecordCollision#` to completely bypass the `LookupRes` constructors? Same for the version in the Strict module.

Data/HashMap/Base.hs
```diff
@@ -751,12 +1083,131 @@ update f = alter (>>= f)
...
+
+alterFWeird
+       :: (Functor f, Eq k, Hashable k)
+       => f (Maybe v)
+       -> f (Maybe v)
+       -> (Maybe v -> f (Maybe v)) -> k -> HashMap k v -> f (HashMap k v)
+alterFWeird _ _ f k m = alterFEager f k m
+{-# INLINE [0] alterFWeird #-}
+
+alterFEager :: (Functor f, Eq k, Hashable k)
```
m-renaud: Can you add a function comment explaining what "eager" means here? Something like:

"This version of alterF always (eagerly) performs the lookup in the map, regardless of if the function f uses the value or not. This is needed so that we can gain information about the position of the key in the map (absent, present with no collision, present with collision at a certain position: see LookupRes) in order to call more specialized versions of insert and delete. We use rewrite rules to select a different implementation when the the function f doesn't care about whether the key is in the map or not, see alterFWeird above."

Data/HashMap/Base.hs
```diff
@@ -1196,18 +1647,25 @@ fromListWith f = L.foldl' (\ m (k, v) -> unsafeInsertWith f k v m) empty
...
+#else
+  forall r k v.
+#endif
+  Eq k => ((# #) -> r) -> (v -> Int -> r) -> k -> A.Array (Leaf k v) -> r
+lookupInArrayCont absent present k0 ary0 = go k0 ary0 0 (A.length ary0)
   where
+    go :: Eq k => k -> A.Array (Leaf k v) -> Int -> Int -> r
     go !k !ary !i !n
-        | i >= n    = Nothing
+        | i >= n    = absent (# #)
```
m-renaud: I'm assuming this doesn't need to be guarded because the earliest compiler version that we build against has support for unboxed tuples, is that correct?

Data/HashMap/Base.hs
m-renaud: No reason. It actually looks like if you do it the way you mentioned you'll save a case statement. Currently we have 3, I think that could be reduced to 2.

Data/HashMap/Strict.hs
treeowl: I doubt you can use them on the left-hand side.

Data/HashMap/Base.hs
treeowl: There's an extra deletion thunk, I believe. I'm not sure we want that.

Data/HashMap/Base.hs
treeowl: I need to check the Core (or you can--start practicing reading `-ddump-simpl`!) but I believe GHC likely unboxes that for us.

treeowl: I need to check the Core (or you can--start practicing reading `-ddump-simpl`!) but I believe GHC likely unboxes that for us.

Data/HashMap/Strict.hs
m-renaud: Yeah, I didn't think so.

Data/HashMap/Base.hs
treeowl: Yes, GHC has had unboxed tuples for many years. I think since the nineties.

Data/HashMap/Base.hs
m-renaud: Learning to read Core is something that's been on my todo list. I'll take a look but it'll probably look like gibberish to me at this point :P

treeowl: It mostly reads like badly formatted Haskell with terrible variable names. Significant differences:

* Type and dictionary arguments are all explicit.

* `case` is *always* strict.

* Recursion is done explicitly with top-level `Rec` or local `letrec`.

* Strict constructors are treated as functions that call the real underlying constructors.

treeowl: For the first while, I recommend using `-dsuppress-all -dno-suppress-type-signatures` to cut down on the clutter. You can reintroduce extra info later.

treeowl: Oh yes, another difference: in the latest versions, you can read `join` and `joinrec` as `let` and `letrec` and ignore `jump`, but know that they're generally more efficient.

m-renaud: Thanks for the info! By "case is always strict" do you mean that the value being `case`d on is always fully evaluated?

treeowl: It's always evaluated to WHNF. Also, the default case is not `_` at the bottom but `DEFAULT` on top, maybe with underscores (don't remember). And the forced value sometimes gets its own name. So Haskell

```haskell
f !x = g x
```

may turn into something like

```
f = \x -> case x of x' { DEFAULT -> g x' }
```

Another thing to note is that there is no infix. Haskell `x +# y` will look in Core like `+# x y`.

treeowl: I've verified that GHC elides the `LookupRes` nonsense as expected. I think it's typically better to limit the ugly hand-unboxing to where we really need it. The only potential exception is the unboxed unary tuple, which is really easy to deal with.

tests/HashMapProperties.hs
treeowl: Actually, I think the rewrites are sufficiently obvious improvements that we shouldn't need to compare. Just need to watch for regressions.

treeowl:
@m-renaud Do you think this is about ready to go?

m-renaud:
I can take a look this afternoon/evening if that's alright.

m-renaud:
Overall the code LGTM. Do you have a recent run of benchmarks comparing the performance of alterF to the specialized functions? I've been busy and haven't had a chance to go through and run the benchmarks myself.

PR #186, opened 2018-02-08
Add keysSet

treeowl:
See #172


treeowl:
I split this into two commits because that's the only way I could find to get git to preserve the history properly, so `git blame` shows the history of `Data.HashMap.Strict.Base` and `Data.HashSet.Base` instead of attributing everthing to my one commit.

PR #190, opened 2018-02-12
Add haddocks for Semigroup and Monoid instances

chris-martin:
Since there is more than one reasonable choice of `mappend` implementation for `HashMap` and `HashSet`, it seems important for there to be some documentation on which one was selected.

### DIFF
--- a/Data/HashMap/Base.hs
+++ b/Data/HashMap/Base.hs
@@ -200,6 +200,8 @@ instance Foldable.Foldable (HashMap k) where
 
 #if __GLASGOW_HASKELL__ >= 711
 -- | '<>' = 'union'
+--
+-- If a key occurs in both maps, the mapping from the first will be the mapping in the result.
 instance (Eq k, Hashable k) => Semigroup (HashMap k v) where
   (<>) = union
   {-# INLINE (<>) #-}
@@ -208,6 +210,8 @@ instance (Eq k, Hashable k) => Semigroup (HashMap k v) where
 -- | 'mempty' = 'empty'
 --
 -- 'mappend' = 'union'
+--
+-- If a key occurs in both maps, the mapping from the first will be the mapping in the result.
 instance (Eq k, Hashable k) => Monoid (HashMap k v) where
   mempty = empty
   {-# INLINE mempty #-}
--- a/Data/HashMap/Base.hs
+++ b/Data/HashMap/Base.hs
@@ -1247,6 +1247,11 @@ alterFEager f !k m = (<$> f mv) $ \fres ->
 
 -- | /O(n+m)/ The union of two maps. If a key occurs in both maps, the
 -- mapping from the first will be the mapping in the result.
+--
+-- ==== __Examples__
+--
+-- >>> union (fromList [(1,'a'),(2,'b')]) (fromList [(2,'c'),(3,'d')])
+-- fromList [(1,'a'),(2,'b'),(3,'d')]
 union :: (Eq k, Hashable k) => HashMap k v -> HashMap k v -> HashMap k v
 union = unionWith const
 {-# INLINABLE union #-}
--- a/Data/HashMap/Base.hs
+++ b/Data/HashMap/Base.hs
@@ -199,11 +199,15 @@ instance Foldable.Foldable (HashMap k) where
     foldr f = foldrWithKey (const f)
 
 #if __GLASGOW_HASKELL__ >= 711
+-- | '<>' = 'union'
 instance (Eq k, Hashable k) => Semigroup (HashMap k v) where
   (<>) = union
   {-# INLINE (<>) #-}
 #endif
 
+-- | 'mempty' = 'empty'
+--
+-- 'mappend' = 'union'
 instance (Eq k, Hashable k) => Monoid (HashMap k v) where
   mempty = empty
   {-# INLINE mempty #-}
--- a/Data/HashMap/Base.hs
+++ b/Data/HashMap/Base.hs
@@ -202,6 +202,11 @@ instance Foldable.Foldable (HashMap k) where
 -- | '<>' = 'union'
 --
 -- If a key occurs in both maps, the mapping from the first will be the mapping in the result.
+--
+-- ==== __Examples__
+--
+-- >>> fromList [(1,'a'),(2,'b')] <> fromList [(2,'c'),(3,'d')]
+-- fromList [(1,'a'),(2,'b'),(3,'d')]
 instance (Eq k, Hashable k) => Semigroup (HashMap k v) where
   (<>) = union
   {-# INLINE (<>) #-}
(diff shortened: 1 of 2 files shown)


Data/HashMap/Base.hs
```diff
@@ -199,11 +199,19 @@ instance Foldable.Foldable (HashMap k) where
     foldr f = foldrWithKey (const f)
 
 #if __GLASGOW_HASKELL__ >= 711
+-- | '<>' = 'union'
+--
+-- If a key occurs in both maps, the mapping from the first will be the mapping in the result.
```
m-renaud: Could you add an example in a collapsible section? You can do that with:

```haskell
-- ==== __Examples__
--
-- example here
--
```

Data/HashMap/Base.hs
```diff
@@ -199,11 +199,19 @@ instance Foldable.Foldable (HashMap k) where
...
 instance (Eq k, Hashable k) => Semigroup (HashMap k v) where
   (<>) = union
   {-# INLINE (<>) #-}
 #endif
 
+-- | 'mempty' = 'empty'
+--
+-- 'mappend' = 'union'
+--
+-- If a key occurs in both maps, the mapping from the first will be the mapping in the result.
```
m-renaud: Example here as well.

Data/HashMap/Base.hs
chris-martin: Shouldn't there be examples in the `union` documentation instead?

m-renaud: Either that with words pointing them to the union for examples, or have the examples in both places.

chris-martin: I guess I can, though I'm confused why that is a blocker to this PR?

m-renaud:
Created #191 to track documentation improvements, this is in the same spirit as https://github.com/haskell/containers/issues/495.

m-renaud:
Not a blocked, just a suggestion if you had the time to add them. If not no
worries, I can do that in a follow up :)

chris-martin:
No worries, examples are a good idea, I'll just go ahead and add them.

Data/HashMap/Base.hs
treeowl: I'm going to disagree with both of you. Long Haddock documentation for instances is (in my experience) always formatted horribly. Therefore, I think the right approach is to limit the documentation string to

```haskell
 +-- | '<>' = 'union' 
```

and let users follow the link to find out what that means.

chris-martin:
I added examples, and also split up the commits so y'all can cherry-pick just the minimal instance documentation if you end up siding with @treeowl.

m-renaud:
Does hiding the examples behind an expandable section address your concern?
The minimal docs you proposed sound fine to me assuming that union has
thorough docs with examples.

Data/HashMap/Base.hs
```diff
@@ -199,11 +199,15 @@ instance Foldable.Foldable (HashMap k) where
     foldr f = foldrWithKey (const f)
 
 #if __GLASGOW_HASKELL__ >= 711
+-- | '<>' = 'union'
 instance (Eq k, Hashable k) => Semigroup (HashMap k v) where
   (<>) = union
   {-# INLINE (<>) #-}
 #endif
 
+-- | 'mempty' = 'empty'
```
treeowl: `mempty` doesn't require documentation unless it's hard to see (e.g., `ZipList`). Speaking of `ZipList`, if you search [`Control.Applicative`](https://hackage.haskell.org/package/base-4.10.1.0/docs/Control-Applicative.html) for `ZipList`, you will see just how absolutely wretchedly horrible non-short Haddocks get for class instances. I therefore think we should probably stick to just `'mappend' = 'union'`. I can edit your commit and merge if you don't want to deal with this any more.

m-renaud: Ew, yeah I didn't realize how bad that is. Also, couldn't the longer `mempty` docs and examples go on the `mempty` definition itself instead of the typeclass instance? Then it would show up under the expandable part of the instance docs and look just like any other function docs.

m-renaud: I would also argue that putting the description to the right of the typeclass name is kinda bizarre, having the description below would probably make longer documentation much more readable, that's a haddock formatting question though.

chris-martin:
My opinion, for what it's worth: Haddock desperately needs to be fixed to legibly display instance documentation, and I'm still holding out hope that the ugly result of libraries actually documenting their instances can put pressure on Haddock to improve.

Data/HashMap/Base.hs
chris-martin: This one I really have to disagree with, because Haskell newbies need all the help they can get. What's obvious to you is not obvious to everyone. The fact that I can type either `mempty` and `empty` and they do exactly the same thing is not universally obvious.

treeowl: @m-renaud, instance methods can't have Haddock documentation, presumably because there's no obvious place to display it. @chris-martin, as best I can tell, the only decent places to write extensive documentation of instances are the introductory section at the very top of the module and in the Haddock documentation for the type itself. I don't have a definitely formed opinion about which of these is less bad.

treeowl: @chris-martin, you may be right that `mempty` is not obvious to all Haskell beginners. But the unfortunate bias in the `HashMap` `Semigroup` instance (that we use `union` rather than `unionWith (<>)`) means that people generally shouldn't use this instance anyway.

m-renaud:
I agree with @chris-martin, the fact that Haddock poorly displays instance documentation isn't a good reason to not provide thorough instance documentation. Maybe this could be done as a follow-up to the Haddock and Hackage [redesign](https://nunoalexandre.com/2018/02/04/redesigning-haskell-docs).

Data/HashMap/Base.hs
treeowl: @m-renaud Hmm.... that looks new! I wonder if we *can* document instance methods now!

treeowl: Changing `(<>)` from `union` to `unionWith (<>)` would require a very, very long process if it were ever undertaken. We'd have to start by having an extensive discussion on the libraries list. Then we'd have to deprecate the `Semigroup` and `Monoid` instances, then remove them. Then we'd have to wait a few more years before putting in the new ones. Then we'd have to cross our fingers and hope no one gets subtle bugs by using old code or reading old documentation. I kind of think we should do it, but it won't actually help anyone till 2025 or something.

chris-martin:
Oh hey, I just had another thought that's kind of related to this.

```
λ> fromList [(1, 2), (1, 3)]
fromList [(1,3)]

λ> foldMap fromList [[(1, 2)], [(1, 3)]]
fromList [(1,2)]
```

We probably should add some docs highlighting the fact that `fromList` prefers the rightmost mapping, but the monoid prefers the leftmost mapping, because that's an easy thing to mix up (I just made a mistake due to misunderstanding this).

treeowl:
I hate these `fromList` functions, and especially `fromListWith`. They're horribly confusing.

m-renaud:
Anyways, @treeowl feel free to merge in whatever you see fit, we can discuss possible changes in a follow-up :)

PR #194, opened 2018-03-16
Make Strict.traverseWithKey actually strict

### LINKED ISSUE #192: Why doesn't 'traverseWithKey' of Data.HashMap.Strict ensure the values of the result in WHNF?
I find a strict version of `traverseWithKey` is missing from Data.HashMap.Strict.Base. Instead I only see following sentence: 
> -- TODO: Should we add a strict traverseWithKey?

Is there any particular reason not to implement it?

I recent chased a memory leak in my program, and eventually find out that the values of my result HashMap after calling traverseWithKey are not in WHNF. This somewhat breaks the meaning of the whole module.

treeowl:
Fixes #192

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,9 @@
 
  * Add `HashMap.keysSet`.
 
+ * Make `HashMap.Strict.traverseWithKey` force the results before
+   installing them in the map.
+
 ## 0.2.9.0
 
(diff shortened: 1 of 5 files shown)


treeowl:
@pierric, do you think you could take a look at this and see if it works for you?

treeowl:
Note: this needs tests before it can be merged.

PR #200, opened 2018-04-20
remove unused array filter function

andrewthad:


treeowl:
Looks reasonable. Speaking of array ops in `unordered-containers`, `Data.HashMap.Base` has a function called `trim` which should be in the array module and should be implemented using `cloneArray#`/`cloneSmallArray#` followed by an unsafe thaw. This should save some useless scribbling time.

andrewthad:
I think you meant that it should be implemented by `cloneMutableArray#` followed by an `unsafeFreezeArray#`. But yes, that would be a good tiny win.

PR #207, opened 2018-04-22
An IVar-based implementation of alterF

### LINKED ISSUE #189: Fix alterFconstant rule
I just realized there's a potential efficiency problem with the `alterFconstant` rule. It works great for `Identity`, `IO`, etc., but it's not so great for something like lists; we lose the location-saving optimization across insertions/deletions. Ugh. What can we do? I think maybe the trick is to get really ugly and use something in the general nature of an `IVar`. That is, go all `unsafePerformIO` on the problem to cache the location.

treeowl:
I haven't benchmarked this, or added even basic rewrite rules,
so it's not really ready to go, but I think the basic idea is
sound.

Note to self: use more `stToIO` to avoid needlessly nested
`runRW#` calls and recover more inlining.

Fixes #189

### DIFF
--- a/Data/HashMap/Array.hs
+++ b/Data/HashMap/Array.hs
@@ -24,6 +24,7 @@ module Data.HashMap.Array
     , indexM
     , index#
     , update
+    , updateM
     , updateWith'
     , unsafeUpdateM
     , insert
--- a/Data/HashMap/Base.hs
+++ b/Data/HashMap/Base.hs
@@ -101,7 +101,8 @@ module Data.HashMap.Base
     , LookupRes(..)
     , insert'
     , delete'
-    , lookup'
+    , insert''
+    , delete''
     , insertNewKey
     , insertKeyExists
     , deleteKeyExists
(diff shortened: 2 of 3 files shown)


treeowl:
@m-renaud and @andrewthad, what do y'all think of this general idea? Do you see any obvious ways to optimize it? It's certainly extreme overkill for `Int` keys, or really any keys that have cheap equality tests, but I haven't been able to think of any other way to both:

1. Avoid performing redundant lookups and
2. Guarantee that a call to `alterF` won't lead to the same key equality tests being repeated.

On the flip side, there's a huge caveat: it's quite possible that this machinery leads to enough of its own overhead that risking redundant lookups is better.

treeowl:
An alternative possibility (considerably more conservative): #208.

treeowl:
I still haven't done any benchmarking, but the more I think about this the less optimistic I get. A `HashMap` lookup is pretty darn cheap when you ignore the cost of hashing and the cost of key equality testing. This version has to allocate an `IORef`, which isn't free. It also allocates a `Present` constructor when appropriate, which *also* isn't free. So this is almost certainly a bad idea.

PR #208, opened 2018-04-22
Restrict alterF constant rule

### LINKED ISSUE #189: Fix alterFconstant rule

treeowl:
The `alterF` rule for recognizing that the given function
ignores its argument can lead to performance problems for
certain functors. Restrict the rule to `Identity`, where
it's always a good idea.

Fixes #189


treeowl:
An alternative direction is #207

PR #217, opened 2018-10-15
WIP: Collisionless containers (DO NOT MERGE)

NorfairKing:


Data/HashMap/Base.hs
```diff
@@ -1822,6 +1799,56 @@ updateOrConcatWithKey f ary1 ary2 = A.run $ do
...
+    Empty -> WasEmpty -- That would be weird, but fine.
+    Leaf _ l -> NowEmpty l
+    BitmapIndexed bm ary' -> case unConsA ary' of
+      WasEmpty -> WasEmpty -- That would be weird, but fine.
+      NowEmpty l -> NowEmpty l
+      UnConsed l a' -> UnConsed l (A.singleton (BitmapIndexed bm a'))
+    Full ary' -> case unConsA ary' of
+        WasEmpty -> WasEmpty -- That would be weird, but fine.
+        NowEmpty l -> NowEmpty l
+        UnConsed l a' -> UnConsed l (A.singleton (Full a')) -- TODO 'Full' seems wrong
```
NorfairKing: This needs careful review.

Data/HashMap/Base.hs
```diff
@@ -1822,6 +1799,56 @@ updateOrConcatWithKey f ary1 ary2 = A.run $ do
...
+  Empty -> Nothing
+  Leaf _ l -> Just (l, Empty)
+  BitmapIndexed bm ary' -> case unConsA ary' of
+    WasEmpty -> Nothing
+    NowEmpty l -> Just (l, Empty)
+    UnConsed l a' -> Just (l, BitmapIndexed bm a')
+  Full ary' -> case unConsA ary' of
+    WasEmpty -> Nothing -- That would be weird, but fine.
+    NowEmpty l -> Just (l, Empty)
+    UnConsed l a' -> Just (l, Full a') -- TODO This 'Full' seems wrong
```
NorfairKing: This needs careful review.

Data/HashMap/Base.hs
```diff
@@ -551,9 +566,9 @@ data LookupRes a = Absent | Present a !Int
 --   Key not in map           => Absent
```
NorfairKing: `LookupRes` probably needs refactoring now.

Data/HashMap/Base.hs
```diff
@@ -1822,6 +1799,56 @@ updateOrConcatWithKey f ary1 ary2 = A.run $ do
...
 {-# INLINABLE updateOrConcatWithKey #-}
 
+-- Helper functions for very rare cases.
+
+-- Remove (any) one element from an array of hashmaps
+unConsA :: A.Array (HashMap k v) -> UnCons k v
+unConsA ary = case A.length ary of
+  0 -> WasEmpty
+  1 -> case A.index ary 0 of
+    Empty -> WasEmpty -- That would be weird, but fine.
```
treeowl: This all looks much too complicated. We maintain the invariant that an array *never* contains an empty map. In addition, this function as currently conceived doesn't seem likely to play well with single-element-array collapsing that I mention below.

Data/HashMap/Base.hs
treeowl: *Both* array cases look deeply suspicious. When a map in a `BitmapIndexed` node becomes empty and is removed, we surely need to update the bitmap. I believe we also want a special case when the array has one remaining element and that element is a leaf: we should surely convert the whole thing to a leaf then. I *believe* we generally maintain the invariant that an array has at least two elements. When a map in a `Full` node goes empty and is removed, we surely need to convert to a `BitmapIndexed` node.

Data/HashMap/Base.hs
```diff
@@ -181,17 +184,17 @@ data HashMap k v
     | BitmapIndexed !Bitmap !(A.Array (HashMap k v))
     | Leaf !Hash !(Leaf k v)
     | Full !(A.Array (HashMap k v))
-    | Collision !Hash !(A.Array (Leaf k v))
+    | Collision !Hash !(Leaf k v) !(Leaf k v) !(HashMap k v)
       deriving (Typeable)
```
treeowl: I think you've missed two invariants (because they're not documented with the type):

1. An array never contains an empty map.
2. An array is never empty.
3. Only a transient array can have exactly one element. These transient arrays are only used to reduce the amount of extra code on collisions.

Data/HashMap/Base.hs
```diff
@@ -698,120 +731,111 @@ insert' h0 k0 v0 m0 = go h0 k0 v0 0 m0
...
-            !st' = go collPos h k x (s+bitsPerSubkey) st
+            !st' = go h s k x (bs+bitsPerSubkey) st
         in Full (update16 ary i st')
-      where i = index h s
-    go collPos h k x _s (Collision _hy v)
-        | collPos >= 0 = Collision h (setAtPosition collPos k x v)
-        | otherwise = Empty -- error "Internal error: go {collPos negative}"
+      where i = index h bs
+    go h s k x bs t@(Collision hx l1@(L k1 v1) l2@(L k2 v2) hmx)
+        | h == hx   =
```
treeowl: The `h == hx` test is guaranteed to succeed in this context because we know the key is in the map.

Data/HashMap/Base.hs
```diff
@@ -1822,6 +1799,56 @@ updateOrConcatWithKey f ary1 ary2 = A.run $ do
...
+    WasEmpty -> Nothing
+    NowEmpty l -> Just (l, Empty)
+    UnConsed l a' -> Just (l, BitmapIndexed bm a')
+  Full ary' -> case unConsA ary' of
+    WasEmpty -> Nothing -- That would be weird, but fine.
+    NowEmpty l -> Just (l, Empty)
+    UnConsed l a' -> Just (l, Full a') -- TODO This 'Full' seems wrong
+  Collision h l1 l2 hm' -> Just $ (,) l1 $ case unConsHM hm' of
+    Nothing -> Leaf h l2
+    Just (l3, hm'') -> Collision h l2 l3 hm''
```
treeowl: This smells very wrong. You're pulling up through a collision level without rehashing. I think the right thing here is probably to special-case `hm' = empty` and in all other cases leave `l1` and `l2` alone and "uncons" from the next level down.

treeowl: Er... Sorry.... Ignore this comment. It's bogus.

NorfairKing: For the record: the comment misses that you don't need to rehash because you already know that the values below will collide on the same hash for this level, because that's the point of the `Collision` constructor.

Data/HashMap/Base.hs
NorfairKing: You're talking about the `A.Array` that I just deleted, right? So these invariants are not necessary anymore?.
Or are you talking about the `BitmapIndexed` constructor?

treeowl: Huh? No... I'm saying you can simplify the code for both `BitmapIndexed` and `Full` by using those invariants, but also that you need to maintain them. Also, if the test suite doesn't check for them, it should--unless I'm wrong and we don't really maintain them.

treeowl:
I stated one invariant incorrectly. We actually *do* allow arrays with exactly one element. Sorry for the confusion. But we *do* (and will continue to) maintain the invariant that every array has at least *one* element, and that none of them ever contain an *empty* element. You will definitely have to do some bitmap updating. I haven't dug into the bitmaps as much as I should have, but in all cases in a `BitmapIndexed bm ar` node, `popCount bm == length ar`. The bitmap indicates which potential 4-bit pieces of hash are actually represented in that array and where each can be found.

Data/HashMap/Base.hs
```diff
@@ -293,8 +297,11 @@ equal1 eq = go
       = bm1 == bm2 && A.sameArray1 go ary1 ary2
     go (Leaf h1 l1) (Leaf h2 l2) = h1 == h2 && leafEq l1 l2
     go (Full ary1) (Full ary2) = A.sameArray1 go ary1 ary2
-    go (Collision h1 ary1) (Collision h2 ary2)
-      = h1 == h2 && isPermutationBy leafEq (A.toList ary1) (A.toList ary2)
+    go (Collision h1 (L k11 v11) (L k12 v12) hm1) (Collision h2 (L k21 v21) (L k22 v22) hm2)
+      = h1 == h2
+      && isPermutationBy (\(k1, v1) (k2, v2) -> k1 == k2 && eq v1 v2)
+           ((k11, v11) : (k12, v12) : toList hm1)
+           ((k21, v21) : (k22, v22) : toList hm2)
```
treeowl: This is not (significantly) worse than what we have now, but it's still O(n^2) when collision buckets are very full. If we have a pathological chain of `Collision` nodes, then we have no hope. But of course that only happens when there are lots of collisions *and* salting doesn't help, which is an exceptionally rare situation. If we add a `Hashable k` constraint here (and therefore to the `Eq` and `Eq1` instances), we should be able to cut it down to linearithmic in the common case where salt works.

```haskell
go t1@(Collision h1 _ _ _) t2@(Collision h2 _ _ _) = h1 == h2 &&
   size t1 == size t2 && foldrWithKey (\k v r -> maybe False (eq v) (lookup k t2) && r) True t1
```

Data/HashMap/Base.hs
```diff
@@ -345,13 +354,15 @@ cmp cmpk cmpv t1 t2 = go (toList' t1 []) (toList' t2 [])
...
-    go (Collision k1 ary1 : tl1) (Collision k2 ary2 : tl2)
-      = compare k1 k2 `mappend`
-        compare (A.length ary1) (A.length ary2) `mappend`
-        unorderedCompare leafCompare (A.toList ary1) (A.toList ary2) `mappend`
+    go (Collision h1 (L k11 v11) (L k12 v12) hm1 : tl1) (Collision h2 (L k21 v21) (L k22 v22) hm2 : tl2)
+      = compare h1 h2 `mappend`
+        compare (length hm1) (length hm2) `mappend`
+        unorderedCompare (\(k1, v1) (k2, v2) -> (k1 `cmpk` k2) `mappend` (v1 `cmpv` v2))
+          ((k11, v11) : (k12, v12) : toList hm1)
+          ((k21, v21) : (k22, v22) : toList hm2) `mappend`
```
treeowl: I'm not at all convinced that `unorderedCompare` is sure to produce a total ordering from two total orderings at all. It looks completely bonkers to me. Unless you can give me a good argument for this function making any sense whatsoever, I think we should *delete* the `Ord2` instance altogether and write a reasonable implementation of `Ord1`.

treeowl: I've asked Oleg if he can explain his reasoning. If it proves solid, we'll have to deal with it, but note in the documentation that `liftCompare2` can have off-the-charts bad performance in pathological cases. Regardless, we should surely consider a simpler and more efficient implementation of the inherently better-behaved `liftCompare`.

Data/HashMap/Base.hs
```diff
@@ -366,9 +377,12 @@ equalKeys1 eq t1 t2 = go (toList' t1 []) (toList' t2 [])
...
       = go tl1 tl2
-    go (Collision k1 ary1 : tl1) (Collision k2 ary2 : tl2)
-      | k1 == k2 && A.length ary1 == A.length ary2 &&
-        isPermutationBy leafEq (A.toList ary1) (A.toList ary2)
+    go (Collision h1 (L k11 _) (L k12 _) hm1 : tl1) (Collision h2 (L k21 _) (L k22 _) hm2 : tl2)
+      | h1 == h2 &&
+        (length hm1 == length hm2) &&
+        isPermutationBy eq
+          (k11 : k12 : keys hm1)
+          (k21 : k22 : keys hm2)
```
treeowl: This is most definitely wrong. You're ignoring the values altogether here!

Data/HashMap/Base.hs
```diff
@@ -385,8 +399,12 @@ equalKeys = go
...
     go (Leaf h1 l1) (Leaf h2 l2) = h1 == h2 && leafEq l1 l2
     go (Full ary1) (Full ary2) = A.sameArray1 go ary1 ary2
-    go (Collision h1 ary1) (Collision h2 ary2)
-      = h1 == h2 && isPermutationBy leafEq (A.toList ary1) (A.toList ary2)
+    go (Collision h1 (L k11 _) (L k12 _) hm1) (Collision h2 (L k21 _) (L k22 _) hm2)
+      = h1 == h2 &&
+        (length hm1 == length hm2) &&
+        isPermutationBy (==)
+          (k11 : k12 : keys hm1)
+          (k21 : k22 : keys hm2)
```
treeowl: Here too, the values are illegitimately ignored.

Data/HashMap/Base.hs
```diff
@@ -401,20 +419,18 @@ instance H.Hashable2 HashMap where
...
-        go s (Collision h a : tl)
-          = (s `H.hashWithSalt` h) `hashCollisionWithSalt` a `go` tl
+        go s (Collision _ l1 l2 hm' : tl)
+          = collisionHash l1 l2 hm' s `go` tl
         go s (_ : tl) = s `go` tl
 
+        collisionHash (L k1 v1) (L k2 v2) hm' s
+                = L.foldl' H.hashWithSalt s
+                $ L.sort . L.map (\(k, v) -> (s `hk` k) `hv` v)
+                $ (k1, v1) : (k2, v2) : toList hm'
```
treeowl: Why do we ignore the hash in the `Collision` node? We used to use it? Perhaps that makes sense for performance, but a comment would be in order.

Data/HashMap/Base.hs
```diff
@@ -428,33 +444,32 @@ instance (Hashable k, Hashable v) => Hashable (HashMap k v) where
...
+toList' Empty                   a = a
 
 -- Helper function to detect 'Leaf's and 'Collision's.
 isLeafOrCollision :: HashMap k v -> Bool
-isLeafOrCollision (Leaf _ _)      = True
-isLeafOrCollision (Collision _ _) = True
-isLeafOrCollision _               = False
+isLeafOrCollision (Leaf _ _)          = True
+isLeafOrCollision (Collision _ _ _ _) = True
+isLeafOrCollision _                   = False
```
treeowl: Why not fix this once and for all?

```haskell
isLeafOrCollision (Leaf {}) = True
isLeafOrCollision (Collision {}) = True
isLeafOrCollision _ = False
```

Data/HashMap/Base.hs
```diff
@@ -483,7 +498,7 @@ size t = go t 0
     go (Leaf _ _)            n = n + 1
     go (BitmapIndexed _ ary) n = A.foldl' (flip go) n ary
     go (Full ary)            n = A.foldl' (flip go) n ary
-    go (Collision _ ary)     n = n + A.length ary
+    go (Collision _ _ _ hm)  n = go hm n + 2
```
treeowl: Tail recursion, please!

```haskell
go (Collision _ _ _ hm)  n = go hm (n + 2)
```

Data/HashMap/Base.hs
treeowl: @phadej has half an explanation and lots of QuickCheck testing. So maybe we can back-burner the general validity question for now. I'm still deeply uncomfortable about the fact that I don't know exactly what relationship must exist among `==`, `hashWithSalt`, and the passed comparison function to make `liftCompare2` behave reasonably.

NorfairKing: @treeowl Looking at the diff between the old code and the new code, I don't think the new code is asymptotically less performant. What's the next action here?

Data/HashMap/Base.hs
NorfairKing: Isn't that the point of `equalKeys`, which has the comment "Same as 'equal' but doesn't compare the values"?
Note that the old and the new semantics are the same.

Data/HashMap/Base.hs
NorfairKing: Same as above. This function is called `equalKeys`.

Data/HashMap/Base.hs
NorfairKing: @treeowl Because we also ignore the hash incase of the `Leaf` constructor. I'm not sure why though. What would you like me to do here?

Data/HashMap/Base.hs
NorfairKing: @treeowl I'm not exactly clear on how the bitmap is supposed to work, but what you say makes complete sense to me.

Data/HashMap/Base.hs
treeowl: You misunderstand. If someone is actually using this instance, and encounters a map with a huge collision bucket, then they will face a problem like the one you're trying to address. I don't think we can fix that for `Ord2` (even though the instance appears to be moderately validish). But we most certainly *can* fix it for `Ord1` and `Ord`, so we should separate those implementations.

Data/HashMap/Base.hs
treeowl: No. `equalKeys` is only for seeing if two things have the same keys (e.g., to implement equality testing for hash *sets*). It has nothing to do with what we're doing *here*.

Data/HashMap/Base.hs
treeowl: I think it's okay for now. The whole thing is a bit weird, but it's not your fault.

Data/HashMap/Base.hs
treeowl: However, there's still something to fix here, I think. Can't we add a `Hashable` constraint and do the same optimization as for real equality?

Data/HashMap/Base.hs
treeowl: Well then, I think you'll have to try to figure that out. You may find it helpful to look at what `delete` does.

Data/HashMap/Base.hs
NorfairKing: @treeowl I can do that for `equalKeys`, but not for `equalKeys1` because we need a `Hashable` constraint, and we don't get that from `Eq1`

treeowl: Right. The `Eq2` and `Ord2` instances for maps and the `Eq1` and `Ord1` instances for sets are hopeless for large collision buckets. That doesn't mean we shouldn't fix what's fixable.

treeowl:
While we're at it, we might as well fix efficiency for `Ord1` (for maps) and `Ord` (for sets).

Data/HashMap/Base.hs
NorfairKing: > That doesn't mean we shouldn't fix what's fixable.

Sorry, I didn't mean to sound like I was suggesting that.

Data/HashMap/Base.hs
```diff
@@ -1822,6 +1800,58 @@ updateOrConcatWithKey f ary1 ary2 = A.run $ do
     return mary
 {-# INLINABLE updateOrConcatWithKey #-}
 
+-- Helper functions for very rare cases.
+
+-- Remove (any) one element from an array of hashmaps
+unConsA :: A.Array (HashMap k v) -> UnCons k v
```
treeowl: Let's start over from scratch here. This all seems much too complicated! I'm thinking something like this:

```haskell
data UnconsHM k v
  = UnconsEmptyHM
  | UnconsedHM !(Leaf k v) !(HashMap k v)

data UnconsA k v
  = NowEmptyA !(Leaf k v)
  | UnconsedFromFirstA !(Leaf k v) !(A.Array (HashMap k v))
  | RemovedFirstA !(Leaf k v) !(A.Array (HashMap k v))

unconsA :: A.Array (HashMap k v) -> UnconsA k v
unconsA ary =
  case unconsHM  (A.index ary 0) of
    UnconsEmptyHM -> error "Unexpectedly empty internal map"
    UnconsedHM lf Empty
      | A.length ary == 1 -> NowEmptyA lf
      | otherwise -> -- Shorten the array and produce RemovedFirstA
    UnconsedHM lf hm' -> -- Replace the 0th element of the array with hm' and produce UnconsedFromFirstA
```

Data/HashMap/Base.hs
```diff
@@ -1822,6 +1800,58 @@ updateOrConcatWithKey f ary1 ary2 = A.run $ do
...
+  deriving (Show)
+
+-- Remove (any) one element from a hashmap
+unConsHM :: HashMap k v -> Maybe (Leaf k v, HashMap k v)
+unConsHM hm = case hm of
+  Empty -> Nothing
+  Leaf _ l -> Just (l, Empty)
+  BitmapIndexed bm ary' -> case unConsA ary' of
+    NowEmpty l -> Just (l, Empty)
+    UnConsed _ l a' -> Just (l, BitmapIndexed bm a') -- TODO ignore the index?
```
treeowl: We need to distinguish between the `UnconsedFromFirstA` and `RemovedFirstA` possibilities. If we `RemovedFirstA`, then we need to update the bitmap.

Data/HashMap/Base.hs
```diff
@@ -1822,6 +1800,58 @@ updateOrConcatWithKey f ary1 ary2 = A.run $ do
...
+  Full ary' -> case unConsA ary' of
+    NowEmpty l -> Just (l, Empty)
+    UnConsed i l a' ->
+      let bm = fullNodeMask .&. complement (1 `unsafeShiftL` i)
+      in Just (l, BitmapIndexed bm a')
```
treeowl: We need to distinguish between the two successful unconsing results here as well. We could still have a `Full` node in the `UnconsedFromFirstA` case. I have not attempted to check your bit arithmetic here yet.

Data/HashMap/Base.hs
```diff
@@ -1594,31 +1629,50 @@ mapMaybe f = mapMaybeWithKey (const f)
...
+      (Just (Leaf _ l1'), Just (Leaf _ l2')) -> Collision h l1' l2' $ go hm
+      (Just (Leaf _ l1'), Nothing) -> case go' hm of
+        Nothing -> Leaf h l1'
+        Just (Leaf _ l3', hm') -> Collision h l1' l3' hm'
+        _ -> error "Should not happen, can be fixed with refactoring I think."
+      (Nothing, Just (Leaf _ l2')) -> case go' hm of
+        Nothing -> Leaf h l2'
+        Just (Leaf _ l3', hm') -> Collision h l2' l3' hm'
+        _ -> error "Should not happen, can be fixed with refactoring I think."
+      (Nothing, Nothing) -> rehash h $ go hm
```
treeowl: Actually, I think I was all wrong about rehashing after all. Very sorry. Let's simplify this:

```haskell
go (Collision h l1 l2 hm) = case (onLeaf (Leaf h l1), onLeaf (Leaf h l2)) of
  (Just (Leaf _ l1'), Just (Leaf _ l2')) -> Collision h l1' l2' $ go hm
  (Just (Leaf _ lf), Nothing) -> go1 lf
  (Nothing, Just (Leaf _ lf)) -> go1 lf
  (Nothing, Nothing) -> case unconsHM (go hm) of
     UnconsEmptyHM -> Empty
     UnconsedHM l1' hm' ->
        case unconsHM hm' of
           UnconsEmptyHM -> Leaf h l1'
           UnconsedHM l2' hm'' -> Collision h l1' l2' hm''

 where
      go1 l1 = case unconsHM (go hm) of
        UnconsEmptyHM -> Leaf h l1
        UnconsedHM l2 hm' -> Collision h l1 l2 hm'
```

Data/HashMap/Base.hs
treeowl: Er... let me adjust that! We probably want to delete from the *last* one rather than the first one, because I imagine that makes the mask update easier.

tests/Validity.hs
```diff
@@ -24,8 +24,9 @@ import qualified Data.HashMap.Lazy as HM
 import qualified Data.HashMap.Array as A
 import Data.HashMap.Base (HashMap(..), Leaf(..), UnconsHM(..))
 import qualified Data.HashMap.Base as HM
-       (bitsPerSubkey, defaultSalt, hashWithSalt, index, mask, nextSalt,
-        sparseIndex, unconsHM)
+       (bitsPerSubkey, cmp1, cmp2, defaultSalt, equal1, equal2,
+        hashWithSalt, index, mask, nextSalt, sparseIndex, unConsA,
+        unConsHM)
```
treeowl: Please import `Data.HashMap.Base` as `HMB` or something for clarity.

tests/Validity.hs
```diff
@@ -270,6 +271,26 @@ pNull = producesValidsOnValids (HM.null :: HM.HashMap Key Int -> Bool)
 pSize :: Property
 pSize = producesValidsOnValids (HM.size :: HM.HashMap Key Int -> Int)
 
+pEqual1 :: Fun (Int, Int) Bool -> Property
+pEqual1 f =
+    producesValidsOnValids2
+        (HM.equal1 (applyFun2 f) :: HM.HashMap Key Int -> HM.HashMap Key Int -> Bool)
```
treeowl: These makes no sense to me whatsoever. Every `Bool` is inherently valid and every `Ordering` is inherently valid. What are you trying to accomplish? I don't think we need these tests at all. We *do* want to add more detailed `Eq2` and `Ord2` tests at some point, but those are going to be on the subtle side. @phadej has promised an explanation of what preconditions those should have, and I'm going to hold him to it. I still hate the way those classes work/not for dictionaries and sets.

tests/Validity.hs
```diff
@@ -89,8 +108,8 @@ instance (Eq k, Hashable k, Validity k, Validity v) =>
                                             check
                                             "The hash, when masked with the current bitmask, of the key, produces the index in the array where the hashmap is found" $
                                         let h = HM.hashWithSalt s k
-                                        in HM.sparseIndex bm (HM.mask h bs) ==
-                                           ix
+                                         in HM.sparseIndex bm (HM.mask h bs) ==
+                                            ix
```
treeowl: Could you indent the `let` and `in` a couple spaces and line them up with each other? I find this hard to read.

tests/Validity.hs
```diff
@@ -114,7 +133,7 @@ instance (Eq k, Hashable k, Validity k, Validity v) =>
                                             check
                                             "The hash, when masked with the current bitmask, of the key, produces the index in the array where the hashmap is found" $
                                         let h = HM.hashWithSalt s k
-                                        in HM.index h bs == ix
+                                         in HM.index h bs == ix
```
treeowl: Same here. I prefer `let` and `in` stacked on top of each other in general.

tests/Validity.hs
```diff
@@ -257,7 +265,7 @@ instance (GenUnchecked k, GenUnchecked v) => GenUnchecked (HashMap k v) where
 
 -- It turns out it's almost impossible to write this instance without using internals.
 -- This is good-enough for now.
-instance (Eq k, Hashable k, GenValid k, GenValid v) =>
+instance (Eq k, Hashable k, GenValid k, Eq v, GenValid v) =>
```
treeowl: This `Eq` looks like an artifact of the temporary change. Probably don't need it?

Data/HashMap/Base.hs
```diff
@@ -261,9 +275,9 @@ instance (Eq k, Hashable k, Read k, Read e) => Read (HashMap k e) where
 
     readListPrec = readListPrecDefault
 
-instance (Show k, Show v) => Show (HashMap k v) where
-    showsPrec d m = showParen (d > 10) $
-      showString "fromList " . shows (toList m)
+-- instance (Show k, Show v) => Show (HashMap k v) where
+--     showsPrec d m = showParen (d > 10) $
+--       showString "fromList " . shows (toList m)
```
treeowl: This instance must be restored!

docs/developer-guide.md
```diff
@@ -111,6 +111,19 @@ updating the spine of the tree).
...
 
+`Collision` is a special case. The previous version used `Collision !Hash
+!(A.Array (HashMap k v))` instead of `Collision !Hash !(Leaf k v) !(Leaf k v)
+!(HashMap k v)`. As a result, hash-collisions would cause `O(number of
+collisions)` insertion time. Because the currently used hash function is not
+collision-resistant, it is cheap to make enough collisions to have this cause a
+denial of service exploit. However, we can use the fact that `hashable` supports
+salted hashing by default to alleviate at least some of this problem.  In the
+current version, insertion time is not `O(number of collisions)` but `O(number
+of consequtive levels of collision)`.  While this is not better in case all the
```
vaibhavsagar: s/consequtive/consecutive/

sjakobi:
Hi @NorfairKing!

Since @emilypi and I have joined @treeowl and @tibbe as maintainers of this package, we've been discussing what to do about the HashDoS potential and what to make of this PR.

I currently think it's very intriguing, others are more sceptical. 

So far the focus of this package and `hashable` has been very much on performance. `hashable` even explicitly recommends against using `HashMap`s for untrusted input:

http://hackage.haskell.org/package/hashable-1.3.0.0/docs/Data-Hashable.html#g:1

But as you are fully aware, some users still use `u-c` with untrusted data.

So if your approach can sufficiently reduce the HashDoS potential without reducing `u-c`'s performance too much and without introducing new pitfalls or security holes, I believe it has a good chance of being merged. After all, a lot of users could benefit!

In any case, we have lots of questions. I'll start with my own:

1. Is this approach your own invention? Has it been tested in other hash map implementations or are there any (reviewed) publications about it?
1. Do you have a practical exploit that is fixed by this PR? If so, please share it with us maintainers privately.
1. How does this PR affect performance? We'll need benchmark results for this, both for "common" inputs, and inputs that could be used for a HashDoS. The most interesting operations to check are probably inserts and lookups. (For comparing the results, I recommend [`criterion-compare`](http://hackage.haskell.org/package/criterion-compare))
1. Is this PR "correct" in that it doesn't introduce any new bugs? Any tests that still need to be added?

sjakobi:
@treeowl There are a lot of open "conversations" resulting from previous reviews of yours. Could you possibly hide those that have been resolved, by clicking on "Resolve conversation" for those?

NorfairKing:
> * Is this approach your own invention? Has it been tested in other hash map implementations or are there any (reviewed) publications about it?

The approach is new-ish. See https://stackoverflow.com/questions/53495133/is-this-approach-to-dealing-with-hash-collisions-new-unique

> * Do you have a practical exploit that is fixed by this PR? If so, please share it with us maintainers privately.

Yes, I have >30K collisions in a ~1MiB json object that can keep a json-parsing server busy for minutes. I've already shared this with tibbe and treeowl.

> * How does this PR affect performance? We'll need benchmark results for this, both for "common" inputs, and inputs that could be used for a HashDoS. The most interesting operations to check are probably inserts and lookups. (For comparing the results, I recommend [`criterion-compare`](http://hackage.haskell.org/package/criterion-compare))

We did the benchmarks as well. If I remember correctly it's in the package that I sent to treeowl and tibbe.

> * Is this PR "correct" in that it doesn't introduce any new bugs? Any tests that still need to be added?

Well the pr needs to be rebased on top of something more recent, but all the tests still passed when I last did that.

sjakobi:
Many thanks for the quick response, @NorfairKing! :)

For now, I'd just like to let you know that our internal discussion is progressing further, and I think we'll be able to give you more feedback within in the next few weeks.

NorfairKing:
> Many thanks for the quick response, @NorfairKing! :)
> 
> For now, I'd just like to let you know that our internal discussion is progressing further, and I think we'll be able to give you more feedback within in the next few weeks.

Thanks for picking this up. I 'm curious to hear more but I am no longer being payed to work on this so I will not be able to justify spending a lot of time on this.

sjakobi:
Out of curiosity: Have you ever considered using [`hashmap`](http://hackage.haskell.org/package/hashmap) instead, @NorfairKing? Since it uses `Data.Map` to store hash collisions, it should be more resistant against HashDoS IMU.

`hashmap` is currently deprecated of course, but it could be [revamped](https://github.com/foxik/hashmap/issues/1).

NorfairKing:
@sjakobi I haven't. We did this work because `aeson` uses `unordered-containers` instead of `containers`, which makes each haskell JSON server vulnerable. So unless you're able to convince the aeson maintainers to break their api and use another map type, it also doesn't matter _to me_.
The collisionless maintainers in this PR fix the problem without breaking API or sacrificing performance.

TomMD:
@NorfairKing Is there a reason this is still WIP and DO NOT MERGE?

@sjakobi The "internal discussion" seems to have gotten lost in the shuffle a year back. A timely fix would be good at this point. Can we merge?

treeowl:
There's still considerable debate about this. Johan Tibell has said no, and
he knows more about this package than anyone.

TomMD:
> Johan Tibell has said no

Is there a link to that discussion so we can hear the objections first hand? As an open source project it would be traditional to have the objection available in the code review.

treeowl:
I'm not sure. There was a lot of private discussion and a lot of things
have fallen out of my cache. I'm sure he'd be happy to comment publicly.

NorfairKing:
> Is there a link to that discussion so we can hear the objections first hand? As an open source project it would be traditional to have the objection available in the code review.

This was a private conversation as part of the responsible disclosure procedure.

Boarders:
@NorfairKing You suggested this doesn't change performance, where are the benchmarks to show that?

NorfairKing:
> @NorfairKing You suggested this doesn't change performance, where are the benchmarks to show that?

Yes, they were included in the original communication, but you can also just try to run the benchmarks locally to see.

Like I said:
         
> We did the benchmarks as well and saw no difference. That makes sense because this implementation only differs when there > are (many) conflicts. If I remember correctly it's in the package that I sent to treeowl and tibbe.

Bodigrim:
@NorfairKing thanks for your work and investigation and this PR, nicely done! 

@TomMD it seems a bit unfair to mount pressure on maintainers of `u-c`: they are responsible neither for the choice of a hash function upstream, nor for downstream packages, which disregarded security warnings in documentation. Performance of `u-c` is a difficult business, and validating such change to the underlying data structure is no small feat. Especially given that this is a novel approach. 

I guess from pragmatic perspective it would be much simpler to replace `Array` by `Map`, but this imposes `Ord`. This is likely a non-issue for mainstream languages, so they never explored approaches avoiding `Ord` such as the proposed sequence of nested `HashMap`s, which I personally find very appealing and beautiful.

sjakobi:
@NorfairKing Thanks for the publishing [this great write-up](https://cs-syd.eu/posts/2021-09-11-json-vulnerability), and apologies for responding so slowly to your PR!

I'm happy that, now that the vulnerability is properly published, we can discuss it more easily in public.

I have opened https://github.com/haskell-unordered-containers/unordered-containers/issues/319, so we can discuss possible responses to the vulnerability separately from the proposed fix that you have provided in this PR. I invite anyone interested to participate in that discussion!

sjakobi:
> @sjakobi The "internal discussion" seems to have gotten lost in the shuffle a year back. A timely fix would be good at this point. Can we merge?

Unfortunately we (at least @tibbe) are not convinced that this PR is of much help with preventing collision attacks. Citing https://github.com/haskell-unordered-containers/unordered-containers/issues/319#issuecomment-918589714:

> @tibbe's assessment was that it would still require a strong hash function and a random seed to be reasonably secure. With many weaker hash functions, it is possible to generate seed-independent collisions, see e.g. [this blog post](https://medium.com/@robertgrosse/generating-64-bit-hash-collisions-to-dos-python-5b21404a5306). By this assessment, #217 "adds" little security of its own.

This is no final judgment on this PR, but I don't currently expect that this PR will be part of a "timely fix" in `u-c`, nor do I currently know of a way to provide a quick fix within `u-c` itself at all.

If you're looking for potential mitigating measures, see https://github.com/haskell-unordered-containers/unordered-containers/issues/319#issuecomment-918267465 and https://github.com/haskell-unordered-containers/unordered-containers/issues/319#issuecomment-918589714.

treeowl:
I agree; this is a fair bit of extra complexity and maintenance burden for something we don't actually know for sure is helpful.

phadej:
> I agree; this is a fair bit of extra complexity and maintenance burden for something we don't actually know for sure is helpful.

Hopefully this a helpful argument: https://medium.com/@robertgrosse/generating-64-bit-hash-collisions-to-dos-python-5b21404a5306 even Robert (the author of that blog post) doesn't share the **universal** collision sets, he is very far in showing they are feasible to construct. FNV-1 is just a weak hash. Thus having hashmap with another hashmap for collision resolution is potentially *very bad*, as that chain simply will not terminate: It's enough to have just two keys which collide universally! Collision resolution method should be vastly different then reusing (weak) hash function.

NorfairKing:
Alright, that convinced me. Closing this for good.

sjakobi:
> Thus having hashmap with another hashmap for collision resolution is potentially _very bad_, as that chain simply will not terminate: It's enough to have just two keys which collide universally!

That's a very good insight @phadej. I hadn't realized this earlier.

Issue #222, opened 2019-02-11
Foldable (HashMap k) is slower than it could be


Fuuzetsu:
Currently the Foldable instance for HashMap only specifies `foldr`. While that's technically enough, it gives poor Core when we use something like `Foldable.foldl'` as it gets translated into a `foldr`, `id` gets passed a few layers down &c.

We could trivially define `Foldable.foldl' = HashMap.foldl'` which is already written, available and generates much better Core. Maybe some other methods should get the same treatment.

Fuuzetsu:
Fixed by #223

treeowl:
We still have some more methods to deal with. I'm working on them.

googleson78:
Is this still active?

I think it would be best to define all the methods (`elem`/`null` are "dangerous" in this aspect, you wouldn't expect them to traverse the structure)

Same thing for `HashSet`

(This would also be nicely consistent with how `containers` defines all the methods)

PR #223, opened 2019-02-11
Extend Foldable (foldl', null, length)

shekeru:
The Foldable instance was slightly extended for issue for Issue #222 out of boredom. Namely adding foldl', null, length, on both HashMap and HashSet. Due to its usage in HashSet, I assumed the `INLINE` pragma was appropriate.

### DIFF
--- a/Data/HashMap/Base.hs
+++ b/Data/HashMap/Base.hs
@@ -201,11 +201,12 @@ instance Foldable.Foldable (HashMap k) where
     {-# INLINE foldr #-}
     foldl' = Data.HashMap.Base.foldl'
     {-# INLINE foldl' #-}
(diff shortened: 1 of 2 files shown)


treeowl:
Why not also `foldl`, `foldr'`, and `foldMap`? While we're at it, we might consider `mapM` and `sequence`.

treeowl:
Er.... No, we can't do anything special for `mapM` or `sequence`. But the other `Foldable` methods would be good.

shekeru:
I assume that would require implementing those functions *somewhere* in the Base.hs files, and since I'm a newbie to Haskell, and with the library overall, I didn't feel comfortable doing that. Sorry for the trouble.

treeowl:
That is true. We'd better do that later.

PR #233, opened 2019-05-06
Make `two` strict in its key arguments

### LINKED ISSUE #232: HashMap has thunks for internal nodes after construction using strict operations
I've been trying to check strictness of my model using `ppHeapTree` function from (slightly altered) ghc-heap-view package and discovered that there seem to be some thunks in HashMap following a series of `insert` using `foldl'`. However if I `union` it with `singleton` it comes out without thunks.

I've create a little project to demonstrate this behaviour:  https://github.com/mantasg/hashmap-test. The version of `ghc-heap-view` I used doesn't work with GHC 8.6 therefore I used stack lts-11.20 with GHC 8.2.  Running it requires `stack` and can be invoked by:
```
cd hashmap-test && stack build hashmap-test && stack exec hashmap-test
```

#### HeapTree for `test1 = foldl' (\m v -> HM.insert v v m) HM.empty [0..5]`
```
HeapTree for test1
"BitmapIndexed ([ (_thunk (I# 0) (I# 0) 0), (_thunk (I# 1) (I# 1) 1), (Leaf (I# 2) (I# 2) 2), (Leaf (I# 3) (I# 3) 3), (Leaf (I# 4) (I# 4) 4), (Leaf (I# 5) (I# 5) 5) ]) 63"
```

#### HeapTree for `test2 = foldl' (\m v -> HM.union (HM.singleton v v) m) HM.empty [0..5]`
```
HeapTree for test2
"BitmapIndexed ([ (Leaf (I# 0) (I# 0) 0), (Leaf (I# 1) (I# 1) 1), (Leaf (I# 2) (I# 2) 2), (Leaf (I# 3) (I# 3) 3), (Leaf (I# 4) (I# 4) 4), (Leaf (I# 5) (I# 5) 5) ]) 63"
```

treeowl:
`two` wasn't strict in its key arguments. We thought this was okay,
because its key arguments are always in WHNF and it's marked
`INLINE`. But `two` is defined as a *recursive* `go` function
(I haven't looked into why), which can't be inlined. I believe
that's the reason GHC doesn't *realize* that the keys are in WHNF.
Anyway, the end result was that `two` would defer the creation of
the `Leaf` values stored in the array, producing very silly thunks.

Fixes #232

### DIFF
--- a/Data/HashMap/Base.hs
+++ b/Data/HashMap/Base.hs
@@ -675,7 +675,7 @@ insert' h0 k0 v0 m0 = go h0 k0 v0 0 m0
                          then t
                          else Leaf h (L k x)
                     else collision h l (L k x)
-        | otherwise = runST (two s h k x hy ky y)
+        | otherwise = runST (two s h k x hy t)
     go h k x s t@(BitmapIndexed b ary)
         | b .&. m == 0 =
             let !ary' = A.insert ary i $! Leaf h (L k x)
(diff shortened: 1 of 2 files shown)


treeowl:
@mantasg, actually, could you test it at the second commit here? I turned things around a bit.

mantasg:
@treeowl, I ran the tests with the changes (at `0d06c107f017937338b9a16aeb73b6e3ae9058ba`) and the `HashMap` comes out without thunks.

treeowl:
Great! I'll merge.

treeowl:
Thanks for bringing this to my attention.

mantasg:
No probs. Thanks for a quick response!

Issue #173, opened 2018-01-05
Add external, beginner friendly docs


m-renaud:
Hey, I recently wrote up some external docs for the `containers` package which [appears to be well received](https://www.reddit.com/r/haskell/comments/7nvjr2/rfc_a_beginner_friendly_introduction_to_haskell/). I was wondering if you would accept a pull request for similar docs for this package. They would be the same format as the `container` docs and would live at haskell-unordered-containers.readthedocs.io.

Let me know what you think!

treeowl:
I think "well-received" is a bit strong. Some have expressed strong
support, some are very much opposed, and some (at least me) support the
idea with significant reservations. I urge you to spend some more time
working out how this idea can work well for one package before expanding to
more. There is still much to be done!

m-renaud:
That's a very reasonable request. Could you elaborate more on what your reservations are? Is it that its not hosted on haskell.org or haskell-lang.org? That it uses ReST instead of haddocks? The styling being different than the other haskell resources?

I have a bunch of ideas around how the documentation story could be improved although I haven't had time to write them up in full yet. Eventually I would love to see every package with API docs (haddocks) at hackage.haskell.org/package/pkg-name, beginner friendly intros at docs.haskell.org/package/pkg-name, and tutorials at docs.haskell.org/tutorials/package/pkg-name.

The reasons I've decided to pursue this using ReadTheDocs because it has a really good "doc writer" experience, the layout is well organized, its easy to change themes (to make the theme align with other haskell docs), also lets you serve the docs from your own domain (making it possible to serve the same docs from docs.haskell.org).

Slightly off topic for here, but a lot of the content in the wikibook seems more suited for this layout as well, and moving it to ReST (or Markdown) and putting it in version control would means it could also be served alongside the other docs.

treeowl:
ReST is not (yet) my favorite thing, but maybe I could learn to like it. My main experience with it has ben muddling through the GHC proposal process. The big points that are still open in my mind:

1. I don't really like to duplicate the basic documentation for a function because that increases the maintenance burden. Is there a way to make snippets of the Haddock documentation appear in the tutorial documentation semi-automatically? That is, in the section of the tutorial describing a function, can you magic in the relevant Haddocks somehow? I don't know much about this web stuff, so maybe it's impossible, but that seems pretty ugly.

[...]

PR #236, opened 2019-05-10
Allow hashable-1.3

phadej:
New major hashable-1.3 have been released:

There are changes which don't manifest in types https://hackage.haskell.org/package/hashable-1.3.0.0/changelog

- Semantic change of Hashable Arg instance to not hash the second argument of Arg in order to be consistent with Eq Arg (#171)
- Semantic change of Hashable Float and Hashable Double instances to hash -0.0 and 0.0 to the same value (#173)
- Add Hashable instance for Fingerprint (#156)
- Add new Data.Hashable.Generic module providing the default implementations genericHashWithSalt and genericLiftHashWithSalt together with other Generics support helpers (#148, #178)
- Bump minimum version requirement of base to base-4.5 (i.e. GHC >= 7.4)

Yet, I don't think they affect `unordered-containers` directly.

(Some *might* depend on `Hashable Float` behavior using e.g. `HashSet`, while not depending on `hashable` directly; but that's nothing `unordered-containers` can do about)

### DIFF
--- a/unordered-containers.cabal
+++ b/unordered-containers.cabal
@@ -43,7 +43,7 @@ library
   build-depends:
     base >= 4.7 && < 5,
     deepseq >= 1.1,
-    hashable >= 1.0.1.1 && < 1.3
+    hashable >= 1.0.1.1 && < 1.4
 
   default-language: Haskell2010
 


phadej:
ping @treeowl

treeowl:
Whoops... Sorry, got distracted looking over the `hashable` changes and forgot to merge. Thanks.

phadej:
@treeowl thanks! Could you make a revision on Hackage, or is there something else pending?

treeowl:
Yes, it's time.

phadej:
I made revisions:
- https://hackage.haskell.org/package/unordered-containers-0.2.10.0/revisions/
- https://hackage.haskell.org/package/unordered-containers-0.2.8.0/revisions/

PR #237, opened 2019-06-23
Document the behavior of fromListWith with non-associative and non-commutative operations

Lysxia:
Also copy the improvements of #133 from Strict to Lazy.

On top of  `f newVal oldVal`, we also need to explain that the elements are accumulated left-to-right. The fact that `fromListWith (++)`  gives the elements in reverse order seems somewhat counter-intuitive to me, but it is what it is.

### DIFF
--- a/Data/HashMap/Base.hs
+++ b/Data/HashMap/Base.hs
@@ -1748,7 +1748,35 @@ fromList = L.foldl' (\ m (k, v) -> unsafeInsert k v m) empty
 {-# INLINABLE fromList #-}
 
 -- | /O(n*log n)/ Construct a map from a list of elements.  Uses
--- the provided function to merge duplicate entries.
+-- the provided function @f@ to merge duplicate entries with
+-- @(f newVal oldVal)@.
+--
+-- === Examples
+--
+-- Given a list @xs@, create a map with the number of occurrences of each
+-- element in @xs@:
+--
+-- > let xs = ['a', 'b', 'a']
+-- > in fromListWith (+) [ (x, 1) | x <- xs ]
+-- >
+-- > = fromList [('a', 2), ('b', 1)]
+--
+-- Given a list of key-value pairs @xs :: [(k, v)]@, group all values by their
+-- keys and return a @HashMap k [v]@.
+--
+-- > let xs = [('a', 1), ('b', 2), ('a', 3)]
+-- > in fromListWith (++) [ (k, [v]) | (k, v) <- xs ]
+-- >
+-- > = fromList [('a', [3, 1]), ('b', [2])]
+--
+-- Note that the lists in the resulting map contain elements in reverse order
+-- from their occurences in the original list.
+--
+-- More generally, duplicate entries are accumulated as follows;
+-- this matters when @f@ is not commutative or not associative.
+--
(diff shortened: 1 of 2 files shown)


emilypi:
For reference, this is what's produced: 

haddocks

Data/HashMap/Base.hs
```diff
@@ -1728,7 +1728,27 @@ fromList = L.foldl' (\ m (k, v) -> unsafeInsert k v m) empty
 {-# INLINABLE fromList #-}
 
 -- | /O(n*log n)/ Construct a map from a list of elements.  Uses
--- the provided function to merge duplicate entries.
+-- the provided function @f@ to merge duplicate entries (@f newVal oldVal@).
```
sjakobi, line 1731: Nit:

```suggestion
-- the provided function @f@ to merge duplicate entries with (@f newVal oldVal@).
```

or even

```suggestion
-- the provided function @f@ to merge duplicate entries with @(f newVal oldVal)@.
```

?

Data/HashMap/Base.hs
```diff
@@ -1728,7 +1728,27 @@ fromList = L.foldl' (\ m (k, v) -> unsafeInsert k v m) empty
...
+-- return a @HashMap k [v]@.
+--
+-- The following example illustrates how duplicate entries are accumulated;
+-- this matters when @f@ is not commutative or not associative.
+--
+-- > fromListWith f [(k, a), (k, b), (k, c), (k, d)] = fromList [(k, f d (f c (f b a)))]
+--
+-- If @f = (++)@ as in the previous example, the lists in the resulting map
+-- contain elements in reverse order from their occurences in the original
+-- list.
```
sjakobi, line 1751: I think a concrete, `doctest-style` example might help get the point across more quickly.

Compare the [corresponding docs for `Data.Map`](http://hackage.haskell.org/package/containers-0.6.2.1/docs/Data-Map-Strict.html#v:fromListWith). Although these suffer from the repeated use of the `'a'` and `'b'` values, which make the issue less obvious.

sjakobi:
BTW, CI has been fixed in the meantime, so a rebase should take care of the red status here.

Data/HashMap/Base.hs
Lysxia: Did you mean showing the output of that expression? It is there on the right, but this is maybe not as clear and visible as it could be.

sjakobi: I think `fromListWith f [(k, a), (k, b), (k, c), (k, d)] = fromList [(k, f d (f c (f b a)))]` is a bit abstract. A concrete example might make the issue a bit clearer.

Lysxia:
Thanks for the reviews :)

Data/HashMap/Base.hs
Lysxia: Oh yes that's true.

Lysxia: Let me do this in this PR too!

Lysxia: Done!

Lysxia:
I made the two existing examples more concrete. The last one is tricky to improve because it would be best illustrated with a non-associative operation, and a good instance is hard to come by.

Lysxia:
Fair enough!

PR #246, opened 2020-01-17
Add 'fromListWithKey' to HashMap

josefs:


treeowl:
Can we reuse the unsafe function you wrote for `unsafeInsertWith`, or are there performance regressions?

josefs:
> Can we reuse the unsafe function you wrote for unsafeInsertWith, or are there performance regressions?

Yes, I suppose we could do that. I haven't looked in to the performance implications yet though. Will do.

josefs:
Btw, the travis failure seems unrelated to my change.

josefs:
Ok, so I've benchmarked the effect of having both full implementations for `unsafeInsertWith` and `unsafeInsertWithKey` compare to defining `unsafeInsertWith` in terms of `unsafeInsertWithKey`. Here are the benchmark for the former:
https://pastebin.com/s0BD8Q1G
And here are the benchmark results for defining `unsafeInsertWith` using `unsafeInsertWithKey`:
https://pastebin.com/X6FbvpZ3

The number changes quite a bit, even for unrelated benchmarks. I assume that's due to code layout effects. If anything, the trend seems to be that it's slightly better define `unsafeInsertWith` using `unsafeInsertWithKey` rather than having two separate implementations. So I'm going going with that. Yell if you disagree @treeowl .

Data/HashMap/Strict/Base.hs
```diff
@@ -615,6 +622,12 @@ fromListWith :: (Eq k, Hashable k) => (v -> v -> v) -> [(k, v)] -> HashMap k v
 fromListWith f = L.foldl' (\ m (k, v) -> unsafeInsertWith f k v m) empty
 {-# INLINE fromListWith #-}
 
+-- | /O(n*log n)/ Construct a map from a list of elements.  Uses
+-- the provided function f to merge duplicate entries (f key newVal oldVal).
```
emilypi, line 626: Docs here could use a bit of help. Suggesting a patch: 

```suggestion
-- the provided function @f@ to merge duplicate entries @(f key newVal oldVal)@.
```

josefs: Done, thanks!

josefs:
I've added some more documentation and tests.

tests/HashMapProperties.hs
```diff
@@ -351,6 +351,12 @@ pFromListWith :: [(Key, Int)] -> Bool
 pFromListWith kvs = (M.toAscList $ M.fromListWith (+) kvs) ==
                     (toAscList $ HM.fromListWith (+) kvs)
 
+pFromListWithKey :: [(Key, Int)] -> Bool
+pFromListWithKey kvs = (M.toAscList $ M.fromListWithKey combine kvs) ==
+                       (toAscList $ HM.fromListWithKey combine kvs)
+  -- Ensure that the way values are combined depend on the key
+  where combine (K k) v1 v2 = (v1 + v2) `mod` (abs k + 1)
```
sjakobi, line 358: I don't find it easy to understand this test. I think `combine` might be commutative though, so it might fail to catch some errors, no?

Data/HashMap/Base.hs
```diff
@@ -1883,6 +1890,33 @@ fromListWith :: (Eq k, Hashable k) => (v -> v -> v) -> [(k, v)] -> HashMap k v
...
 {-# INLINE fromListWith #-}
 
+-- | /O(n*log n)/ Construct a map from a list of elements.  Uses
+-- the provided function to merge duplicate entries.
+--
+-- === Examples
+--
+-- Given a list of key-value pairs where the keys are of different flavours, e.g:
+--
+-- > data Key = Mul | Add
```
sjakobi, line 1900: I think the result of the example below depends on the Hashable instance, so you should probably include it here.

Data/HashMap/Base.hs
```diff
@@ -1883,6 +1890,33 @@ fromListWith :: (Eq k, Hashable k) => (v -> v -> v) -> [(k, v)] -> HashMap k v
...
+--
+-- and the values need to be combined differently on a collision, depending on the key:
```
sjakobi, line 1902: `collision` sounds like hash collision, but I don't think that's what you mean.

Data/HashMap/Base.hs
```diff
@@ -1883,6 +1890,33 @@ fromListWith :: (Eq k, Hashable k) => (v -> v -> v) -> [(k, v)] -> HashMap k v
...
+--
+-- > combine Mul = (*)
+-- > combine Add = (+)
```
sjakobi, line 1905: Commutative operations make it harder to infer the behaviour of the function from the example.

tests/HashMapProperties.hs
josefs: I agree it's a bit contrived. But it's not that much different from the test above it, it's just the `combine` function which mixes in the key in the computation.

As for commutativity, the `pFromListWith` uses commutative operators and there's where I got the inspiration. I suppose I could use e.g. subtraction instead.

sjakobi: Yeah, feel free to improve `pFromListWith` too! Subtraction is both non-commutative _and_ non-associative, which seems ideal!

Data/HashMap/Base.hs
josefs: Yes indeed, the result depends on the `Hashable` instance. But I left it out because I felt that it was a distraction and didn't really add anything important. There are only two instances here, one which will give you the order that I provided in the example, and the other which will give you the reverse order. Since the order isn't really the point of the example I'd rather leave out the instance.

Data/HashMap/Base.hs
josefs: Good point. I'll update the docs.

Data/HashMap/Base.hs
josefs: Really? I thought that they would make it easier! The example is supposed to use the difference between addition and multiplication as a way to let the key affect combining the keys.

What would you suggest I use instead?

I suppose I could use something like taking the first or the last element, depending on the key.

sjakobi: Maybe something like

    combine k x y = [ k ] ++ x ++ y

?

With a commutative operation you can't tell in which order the elements are being inserted which is an important gotcha here.

tests/HashMapProperties.hs
josefs: Done. Thanks.

Data/HashMap/Base.hs
josefs: I'm not a big fan of combining the key with the values in the way you suggest. That's how it's documented in Data.Map and I specifically tried to avoid it because I think it seems very contrived. 

The point about the order of operations is made in the last lines of the documentation. This mirrors the documentation of `fromListWith` which also has examples which uses commutative operators but then explains below the order of operations. I kind of like that structure.

treeowl:
Free magmas are great for arbitrary operations.

```haskell
data Magma a
  = Leaf a
  | Op (Magma a) (Magma a)
```

If you pull up an arbitrary `HashMap a` and `fmap Leaf` over it, then you can use `Op` as your combining function. `Op` has no structure at all, so any mistake will not get cancelled out anywhere else.

Data/HashMap/Base.hs
sjakobi: Hmm, the thing is that IMHO `fromListWith` and `fromListWithKey` really have this rather serious design mistake where

    fromListWith op [(k, a), (k, b)]

is not the obvious

    fromList [(k, op a b)]

but instead

    fromList [(k, op b a)]

I wish we could fix this properly but that's not a task for this PR.

So IMHO we need to ensure that potential users of `fromListWith[Key]` become aware of this gotcha ASAP so they _do not_ produce a bug, for which I feel we, as the library authors and maintainers, would share some responsibility!

Currently the only way these haddocks _hint_ at the gotcha is in the last two lines, which IMHO are fairly abstract and easy to overlook TLDR-style.

I think the documentation for `fromListWith` is less problematic because the first example with the commutative operation is very short, and the next example is very clear about the issue.

So maybe consider using David's free magma suggestion (which I feel is _slightly_ abstract) or maybe just use `Sub` and `Div` instead of `Add` and `Mul`…

treeowl:
Oh, I got mixed up. Thought you were talking about testing rather than
documentation. The ergonomics are indeed quite awful. It would be much
clearer if we had something like

flwk
  :: (Hashable k, Eq k)
  => (k -> Maybe b -> a -> Box b) -> [(k,a)] -> HashMap k b

(where data Box a = Box a)

or even

flwkm
  :: (Hashable k, Eq k)
  => (k -> Maybe b -> a -> Maybe b) -> [(k,a)] -> HashMap k b

sjakobi:
@treeowl Yeah, we also discussed tests. There, I'm satisfied with the use of `(-)` as a non-commutative and non-associative operation.

@josefs You could of course consider using the same or a similar function for the example too!

I've opened https://github.com/haskell-unordered-containers/unordered-containers/issues/264 to dicuss the `fromListWith` ergonomics.

sjakobi:
@josefs I would like to include this PR in the upcoming release (see https://github.com/haskell-unordered-containers/unordered-containers/pull/266). Do you think you could address the outstanding documentation issue within the next 7 days?

@treeowl I assume you have vetted the internals and are happy with them!?

josefs:
@sjakobi I've updated the documentation to use division and subtraction instead to better illustrate the non-obvious way in which values are combined.

I really liked the free magma suggestion from @treeowl so I've converted the tests to use that.

I believe that should take care of all the outstanding issues for this PR so it should be good to go for the next release.

Btw, it might be good to squash at least some of these commits.

PR #252, opened 2020-04-21
Add a call stack to (!)

feuerbach:
This adds a helpful stack trace to the `Data.HashMap.Base.(!): key not found` message, allowing the programmer to locate easily the failing call to `(!)`.


emilypi:
`HasCallStack` is usually pretty useful and relatively painless. 👍 from me

nh2:
This comes just at the right time, thanks!

PR #258, opened 2020-05-31
#254, fix space leak on collisions

### LINKED ISSUE #254: Space leak with collisions
Given the following program:

```
import qualified Data.HashMap.Strict as Map
import Data.IORef
import Data.Hashable
import Control.Monad

newtype Wrapper a = Wrapper a
    deriving Eq

instance Hashable (Wrapper a) where
    hashWithSalt _ _ = 0

main = do
    mp <- newIORef $ Map.fromList [(Wrapper 0, "0"), (Wrapper 1, "1")]
    forM_ [1..1000000000] $ \i -> do
        modifyIORef' mp $ Map.insert (Wrapper 2) (show i)
```

It leaks about 1Gb/s when compiled with `-O`. It is an example cut down from Ghcide, https://github.com/digital-asset/ghcide/pull/586, where it leaks 30Mb/s. The issue is that `updateOrSnocWith` lazily applies a function to the key, the old value, and the new value. In the case of inserting with a collision, the only relevant values are the old values, and everything else is `const` ignored. One remedy, at the cost of duplicating code, so to write a dedicated `updateOrSnoc` that skips the function call entirely, as anything else is likely to overly-force values in the .Lazy case.

ndmitchell:
Fixes #254. That benchmark used to leak 1Gb/s, now it doesn't leak and runs in 5Mb.

I added a test case. The invariants to ensure GHC doesn't hoist the memory out (and thus keep it alive) are quite subtle, but the result should be fairly robust. I confirm the test fails before, and passes after. Unfortunately, it's not as simple as testing for strictness, since the functions have the right strictness behaviour, it's just old strict values get retained longer than is necessary.


tests/Regressions.hs
```diff
@@ -71,6 +76,48 @@ propEqAfterDelete (Keys keys) =
...
+  hashWithSalt salt _ = salt
+
+issue254Lazy :: Assertion
+issue254Lazy = issue254LazyLambda 2
+
+-- Important that oldV is not hoisted out by optimisation, so use NOINLINE
+{-# NOINLINE issue254LazyLambda #-}
+issue254LazyLambda :: Int -> Assertion
+issue254LazyLambda i = do
+  _ <- return ()
```
sjakobi, line 94: What does this line do?

tests/Regressions.hs
```diff
@@ -71,6 +76,48 @@ propEqAfterDelete (Keys keys) =
 mapFromKeys :: [Int] -> HM.HashMap Int ()
 mapFromKeys keys = HM.fromList (zip keys (repeat ()))
 
+------------------------------------------------------------------------
+-- Issue #254
+
+data KC = KC Int
```
sjakobi, line 82: Is `KC` an abbreviation? For what?

Data/HashMap/Base.hs
```diff
@@ -1932,12 +1932,12 @@ updateWith# f k0 ary0 = go k0 ary0 0 (A.length ary0)
...
 {-# INLINABLE updateWith# #-}
 
-updateOrSnocWith :: Eq k => (v -> v -> v) -> k -> v -> A.Array (Leaf k v)
+updateOrSnocWith :: Eq k => (v -> v -> (# v #)) -> k -> v -> A.Array (Leaf k v)
                  -> A.Array (Leaf k v)
 updateOrSnocWith f = updateOrSnocWithKey (const f)
 {-# INLINABLE updateOrSnocWith #-}
 
-updateOrSnocWithKey :: Eq k => (k -> v -> v -> v) -> k -> v -> A.Array (Leaf k v)
+updateOrSnocWithKey :: Eq k => (k -> v -> v -> (# v #)) -> k -> v -> A.Array (Leaf k v)
```
sjakobi, line 1940: Maybe add a reference to David's explanation of the unboxed-tuple trick here?

tests/Regressions.hs
```diff
@@ -71,6 +76,48 @@ propEqAfterDelete (Keys keys) =
...
+issue254LazyLambda :: Int -> Assertion
+issue254LazyLambda i = do
+  _ <- return ()
+  let oldV = show i
+  weakV <- mkWeakPtr oldV Nothing
+  let mp = HML.insert (KC 1) "3" $ HML.fromList [(KC 0, "1"), (KC 1, oldV)]
+  _ <- evaluate mp
+  performGC
+  res <- deRefWeak weakV
+  _ <- evaluate mp
```
sjakobi, line 101: What does this second `evaluate` do?

tests/Regressions.hs
sjakobi: Is it important that this is a `data` type and not a newtype?

tests/Regressions.hs
```diff
@@ -71,6 +76,48 @@ propEqAfterDelete (Keys keys) =
 mapFromKeys :: [Int] -> HM.HashMap Int ()
 mapFromKeys keys = HM.fromList (zip keys (repeat ()))
 
+------------------------------------------------------------------------
+-- Issue #254
+
+data KC = KC Int
+  deriving (Eq, Ord, Show)
+instance Hashable KC where
+  hashWithSalt salt _ = salt
```
sjakobi, line 85: The laziness in the `KC` value is completely unimportant, right? The only relevant space leak is  the one in `updateSnocWithKey`?!

tests/Regressions.hs
```diff
@@ -71,6 +76,48 @@ propEqAfterDelete (Keys keys) =
 mapFromKeys :: [Int] -> HM.HashMap Int ()
 mapFromKeys keys = HM.fromList (zip keys (repeat ()))
 
+------------------------------------------------------------------------
+-- Issue #254
```
sjakobi, line 80: I've been able to reproduce the issue on my local machine by running these tests without your fix. Amazing! :)

These tests use (Simon Marlow's ?) weak-reference-based space leak detection technique, right? I've never actually seen it used in practice!

Could you possibly add a reference to an explanation of the technique?

tests/Regressions.hs
ndmitchell: KeyCollides. Data vs newtype is irrelevant. I updated the test to say that.

tests/Regressions.hs
ndmitchell: Puts the let binding under a lambda, so it doesn't get hoisted out and shared in the root. Super subtle. I updated the comments to say so.

Data/HashMap/Base.hs
ndmitchell: We use the unboxed tuple trick in a bunch of places in HashMap, so it seems "expected knowledge" for working with this package.

tests/Regressions.hs
ndmitchell: Makes sure the map itself is retained after the GC. I added a comment.

tests/Regressions.hs
ndmitchell: Yes.

tests/Regressions.hs
ndmitchell: Done.

ndmitchell:
Lots more documentation about the test based on the review by @sjakobi. I also made the lazy test use undefined values to insert, checking that you don't have the space leak but still aren't strict.

treeowl:
Will look today.

Data/HashMap/Base.hs
emilypi: Perhaps we should raise a general doc issue for this trick then. It's uncommon knowledge I think 😄

emilypi:
So, I'm 👍 on this - I can't find anything in particular to gripe about. If @treeowl sees something to fix, i think he should be the final approval here.

Data/HashMap/Base.hs
ndmitchell: I think a blog post explaining the trick would be great. I wouldn't document it in the context of unordered-containers, but try and make it a general technique, and just link to it from this project.

sjakobi: Yeah, a blog post would be great. I have opened https://github.com/haskell-unordered-containers/unordered-containers/issues/262 as a reminder.

Data/HashMap/Base.hs
```diff
@@ -1949,7 +1949,7 @@ updateOrSnocWithKey f k0 v0 ary0 = go k0 v0 ary0 0 (A.length ary0)
             A.write mary n (L k v)
             return mary
         | otherwise = case A.index ary i of
-            (L kx y) | k == kx   -> A.update ary i (L k (f k v y))
+            (L kx y) | k == kx, (# v2 #) <- f k v y -> A.update ary i (L k v2)
```
sjakobi, line 1952: I'm still a bit mystified how the fix works, but this must be the crucial line.

By forcing the unboxed tuple, we can get rid of the reference to the old value `y`, so it can be GC'd. Is that right?

ndmitchell: This line does the forcing, so it's half the story. Line 760 constructs `\a _ -> (# a #)` which says how far to evaluate (at least as far as that the `_` has disappeared). Together they get rid of the old value, that previously hung out in the `_`, and is now not in the `(# #)` value.

Data/HashMap/Base.hs
treeowl: I would call this `updateOrSnocWithKey#`, and define `updateOrSnocWithKey` to use it in the boring way. As it is, this function can only be used for the *lazy* operations. Should we extend it to let it work for strict ones too? The most semantically pleasing version would look like this:

```haskell
updateOrSnocWith#
  :: Eq k
  => ((# (# #) | v #) -> (# v #))
  -> k
  -> A.Array (Leaf k v)
  -> A.Array (Leaf k v)
  -> A.Array (Leaf k v)
```

This takes a function that we pass either `(# (# #) | #)` (meaning the key was not present) or `(# | v #)` (the value tied to the key in the map) and produces an unboxed unary tuple for the new value.

Unfortunately, we support GHC versions that don't have unboxed sums. We can work around that in one of two ways:

```haskell
updateOrSnocWith#
  :: Eq k
  => (Maybe v -> (# v #))
  -> k
  -> A.Array (Leaf k v)
  -> A.Array (Leaf k v)
  -> A.Array (Leaf k v)
```

```haskell
updateOrSnocWith#
  :: Eq k
  => (v -> (# v #))
  -> k
  -> ((# #) -> (# v #))
  -> A.Array (Leaf k v)
  -> A.Array (Leaf k v)
  -> A.Array (Leaf k v)
```

Another issue is evaluating the whole `v -> v` vs. `v -> v -> v` sort of approach, which has to do with what closures we might have to create in different contexts. This will come down to poring over Core and doing some benchmarking. This is actually true whether we try to unify strict and lazy or not. Personally, I'd *love* to be able to unify as much as we can, if we can get enough of the right things to inline the way we want to avoid hurting performance.

Final note: if we're careful enough about inlining, we might well be able to just use boxed everything and still not pay for it. That means making sure that the underlying implementation inlines into the `Strict` or `Lazy` version as soon as the wrapped-up function is passed in, while that `Strict` or `Lazy` version itself remains (at least) `INLINABLE`.

ndmitchell: Is it useful to add both `updateOrSnocWith#` and `updateOrSnocWith`? For user code, where we expose it, I agree that having a "friendly" wrapper is valuable. But given we have `updateOrSnocWith` and `updateOrSnocWithKey` that seems like a lot of intermediate helper functions that don't get used a whole lot.

My inclination would be to keep this patch (which fixes a space leak) a small self-contained space leak fix, and then follow up with the larger unboxed sums and looking at the Core after. It feels like a subsequent refactor, and one that I'm not well placed to make.

Data/HashMap/Base.hs
treeowl: Speaking of this line, that case-plus guard can be rewritten

```haskell
        | L kx y <- A.index ary i
        = if
             | k == kx
             , (# v2 #) <- f k v y
             -> A.update ary i (L k v2)
             | otherwise
             -> go k v ary (i+1) n
```

Personally, I think this seems a little easier to follow.

tests/Regressions.hs
```diff
@@ -71,6 +76,57 @@ propEqAfterDelete (Keys keys) =
...
+issue254Lazy = issue254LazyLambda 2
+
+-- We want to make sure that old values in the HashMap are evicted when new values are inserted,
+-- even if they aren't evaluated. To do that, we use the WeakPtr trick described at
+-- http://simonmar.github.io/posts/2018-06-20-Finding-fixing-space-leaks.html.
+-- We insert a value named oldV into the HashMap, then insert over it, checking oldV is no longer reachable.
+--
+-- To make the test robust, it's important that oldV isn't hoisted up to the top or shared. To do that,
+-- we use NOINLINE, make oldV dependent on an unseen argument, and insert _ <- return () to ensure oldV
+-- is under a lambda.
```
treeowl, line 98: Yuck yuck yuck. I don't like this approach to avoiding floating. It seems much too fragile. Fortunately, I think we can do something both simpler and more robust, taking advantage of the surprising fact that `type Assertion = IO ()`.

```haskell
issue254Lazy :: Assertion
issue254Lazy = do
  i :: Int <- randomIO
  let oldV = error $ "...." ++ show i
  ....
```

Now GHC has no way to float out the old value because it's genuinely unknown until the assertion is run.

tests/Regressions.hs
```diff
@@ -71,6 +76,57 @@ propEqAfterDelete (Keys keys) =
...
+{-# NOINLINE issue254LazyLambda #-}
+issue254LazyLambda :: Int -> Assertion
+issue254LazyLambda i = do
+  _ <- return () -- put oldV under a lambda
+  let oldV = error $ "Should not be evaluated: " ++ show i
+  weakV <- mkWeakPtr oldV Nothing -- test whether oldV is alive
+  let mp = HML.insert (KC 1) (error "Should not be evaluated") $ HML.fromList [(KC 0, "1"), (KC 1, oldV)]
+  _ <- evaluate mp -- force the insert to happen
```
treeowl, line 106: This should probably be written

```haskell
  mp <- evaluate $ HML.insert (KC 1) (error "Should not be evaluated") $ HML.fromList [(KC 0, "1"), (KC 1, oldV)]
```

tests/Regressions.hs
```diff
@@ -71,6 +76,57 @@ propEqAfterDelete (Keys keys) =
...
+  performGC
+  res <- deRefWeak weakV -- gives Just if oldV is still alive
+  _ <- evaluate mp -- makes sure that we didn't GC away the whole HashMap, just oldV
```
treeowl, line 109: Don't `evaluate mp` here; that doesn't really explain our aim, and in principle GHC could simply erase it because `mp` has already been evaluated. What we really want to do is `touch mp`.

```haskell
import GHC.Exts ( touch# )
import GHC.IO (IO (..))
touch :: a -> IO ()
touch a = IO (\s -> (# touch# a s, () #))
```

The very latest GHC version has a new `with#` function (not yet a primop) that's more reliable than `touch#`, but I don't think anything we're doing here could trip the flakiness in `touch#`.

tests/Regressions.hs
ndmitchell: That's way better. I've done that.

tests/Regressions.hs
ndmitchell: Done.

tests/Regressions.hs
ndmitchell: My desire to reach into the guts of the IO monad is low - I think GHC really should expose `touch` in Control.Exception itself. But in the absence of that, I've copied the code above.

treeowl:
I have no idea why `touch` is not available in `System.Mem.Weak`. That seems rather peculiar. Hopefully the new `with` will be exposed there at some point.

Data/HashMap/Base.hs
ndmitchell: I refactored to morally follow that pattern - the nested `if` is not actually needed as the `kx` and `y` bound by `A.index` aren't used (and importantly, _shouldn't_ be used) in the `go` recursive call - so I hoisted it all up a bit.

unordered-containers.cabal
```diff
@@ -154,6 +154,7 @@ test-suite regressions
     base,
     hashable >= 1.0.1.1,
     HUnit,
+    random,
     QuickCheck >= 2.4.0.1,
```
sjakobi, line 158:
```suggestion
    QuickCheck >= 2.4.0.1,
    random,
```

treeowl:
Thanks!

treeowl:
I want to add extra special thanks for the testing work. That was all sorts of tricky and subtle.

PR #249, opened 2020-03-26
Create ReadTheDocs project for unordered-containers.

m-renaud:
These docs walk a newcomer through the unordered-containers package.

Content includes:
- Overview of what's available in the package
- Related packages
- Links to more resources
- How to import and add a dependency on the package
- Walk through common functions from HashSet and HashMap

Rendered docs can be found at https://m-renaud-haskell-unordered-containers.readthedocs.io/en/m-renaud-docs-user-guide/

Once approved these will be published at https://haskell-unordered-containers.readthedocs.io.

This resolves #173.  

/cc @treeowl for review

PR #267, opened 2020-06-08
WIP: Improve unordered-containers API docs.

m-renaud:
This integrates the content from #249 into the module docs themselves instead of having a separate Tutorial module.

Changes:
- Improve top level module docs
- Include examples for all functions
- Call out complexity separately from function summary

### DIFF
--- a/Data/HashSet.hs
+++ b/Data/HashSet.hs
@@ -4,25 +4,94 @@
 #endif
 
 ------------------------------------------------------------------------
--- |
--- Module      :  Data.HashSet
--- Copyright   :  2011 Bryan O'Sullivan
--- License     :  BSD-style
--- Maintainer  :  johan.tibell@gmail.com
--- Stability   :  provisional
--- Portability :  portable
---
--- A set of /hashable/ values.  A set cannot contain duplicate items.
--- A 'HashSet' makes no guarantees as to the order of its elements.
---
--- The implementation is based on /hash array mapped trie/.  A
--- 'HashSet' is often faster than other tree-based set types,
--- especially when value comparison is expensive, as in the case of
--- strings.
---
--- Many operations have a average-case complexity of /O(log n)/.  The
--- implementation uses a large base (i.e. 16) so in practice these
--- operations are constant time.
+{-|
+Module      :  Data.HashSet
+Copyright   :  2011 Bryan O'Sullivan
+License     :  BSD-style
+Maintainer  :  johan.tibell@gmail.com
+Stability   :  provisional
+Portability :  portable
+
+= Introduction
+
+'HashSet' allows you to store /unique/ elements, providing efficient insertion,
+lookups, and deletion. A 'HashSet' makes no guarantees as to the order of its
+elements.
+
+If you are storing sets of "Data.Int"s consider using "Data.IntSet" from the
+<https://hackage.haskell.org/package/containers containers> package.
+
+
+== Examples
+
+All the examples below assume @HashSet@ is imported qualified, and uses the following @dataStructures@ set.
+
+>>> import qualified Data.HashSet as HashSet
+>>> let dataStructures = HashSet.fromList ["Set", "Map", "Graph", "Sequence"]
+
+=== Basic Operations
+
+Check membership in a set:
+
+>>> -- Check if "Map" and "Trie" are in the set of data structures.
+>>> HashSet.member "Map" dataStructures
+True
+>>> HashSet.member "Trie" dataStructures
+False
+
+Add a new entry to the set:
+
+>>> let moreDataStructures = HashSet.insert "Trie" dataStructures
+>>> HashSet.member "Trie" moreDataStructures
+> True
+
+Remove the @\"Graph\"@ entry from the set of data structures.
+
+>>> let fewerDataStructures = HashSet.delete "Graph" dataStructures
+>>> HashSet.toList fewerDataStructures
+["Map","Set","Sequence"]
+
+
+Create a new set and combine it with our original set.
+
+>>> let unorderedDataStructures = HashSet.fromList ["HashSet", "HashMap"]
+>>> HashSet.union dataStructures unorderedDataStructures
+fromList ["Map","HashSet","Graph","HashMap","Set","Sequence"]
+
+=== Using custom data with HashSet
+
+To create a @HashSet@ of your custom type, the type must have instances for
+'Data.Eq.Eq' and 'Data.Hashable.Hashable'. The @Hashable@ typeclass is defined in the
+<https://hackage.haskell.org/package/hashable hashable> package, see the
+documentation for information on how to make your type an instance of
+@Hashable@.
+
+We'll start by setting up our custom data type:
+
+>>> :set -XDeriveGeneric
+>>> import GHC.Generics (Generic)
+>>> import Data.Hashable
+>>> data Person = Person { name :: String, likesDogs :: Bool } deriving (Show, Eq, Generic)
+>>> instance Hashable Person
+
+And now we'll use it!
+
+>>> let people = HashSet.fromList [Person "Lana" True, Person "Joe" False, Person "Simon" True]
+>>> HashSet.filter likesDogs people
+fromList [Person {name = "Simon", likesDogs = True},Person {name = "Lana", likesDogs = True}]
+
+
+== Performance
+
+The implementation is based on /hash array mapped tries/.  A
+'HashSet' is often faster than other 'Data.Ord.Ord'-based set types,
+especially when value comparisons are expensive, as in the case of
+strings.
+
+Many operations have a average-case complexity of /O(log n)/.  The
+implementation uses a large base (i.e. 16) so in practice these
+operations are constant time.
+-}
 
 module Data.HashSet
     (
--- a/Data/HashSet/Base.hs
+++ b/Data/HashSet/Base.hs
@@ -19,7 +19,7 @@
 -- A set of /hashable/ values.  A set cannot contain duplicate items.
 -- A 'HashSet' makes no guarantees as to the order of its elements.
 --
--- The implementation is based on /hash array mapped trie/.  A
+-- The implementation is based on /hash array mapped tries/.  A
 -- 'HashSet' is often faster than other tree-based set types,
 -- especially when value comparison is expensive, as in the case of
 -- strings.
@@ -36,10 +36,6 @@ module Data.HashSet.Base
     , empty
     , singleton
 
-    -- * Combine
-    , union
-    , unions
-
     -- * Basic interface
     , null
     , size
@@ -49,6 +45,10 @@ module Data.HashSet.Base
 
     -- * Transformations
     , map
+
+    -- * Combine
+    , union
+    , unions
 
       -- * Difference and intersection
     , difference
@@ -260,23 +260,38 @@ hashSetDataType :: DataType
 hashSetDataType = mkDataType "Data.HashSet.Base.HashSet" [fromListConstr]
 
 -- | /O(1)/ Construct an empty set.
+--
+-- >>> HashSet.empty
+-- fromList []
 empty :: HashSet a
 empty = HashSet H.empty
 
 -- | /O(1)/ Construct a set with a single element.
+--
+-- >>> HashSet.singleton 1
+-- fromList [1]
 singleton :: Hashable a => a -> HashSet a
 singleton a = HashSet (H.singleton a ())
 {-# INLINABLE singleton #-}
 
--- | /O(1)/ Convert to the equivalent 'HashMap'.
+-- | /O(1)/ Convert to set to the equivalent 'HashMap' with @()@ values.
+--
+-- >>> HashSet.toMap (HashSet.singleton 1)
+-- fromList [(1,())]
 toMap :: HashSet a -> HashMap a ()
 toMap = asMap
 
--- | /O(1)/ Convert from the equivalent 'HashMap'.
+-- | /O(1)/ Convert from the equivalent 'HashMap' with @()@ values.
+--
+-- >>> HashSet.fromMap (HashMap.singleton 1 ())
+-- fromList [1]
 fromMap :: HashMap a () -> HashSet a
 fromMap = HashSet
 
 -- | /O(n)/ Produce a 'HashSet' of all the keys in the given 'HashMap'.
+--
+-- >>> HashSet.keysSet (HashMap.fromList [(1, "a"), (2, "b")]
+-- fromList [1,2]
 --
 -- @since 0.2.10.0
 keysSet :: HashMap k a -> HashSet k
@@ -286,8 +301,6 @@ keysSet m = fromMap (() <$ m)
 --
 -- To obtain good performance, the smaller set must be presented as
 -- the first argument.
---
--- ==== __Examples__
 --
 -- >>> union (fromList [1,2]) (fromList [2,3])
 -- fromList [1,2,3]
@@ -303,17 +316,32 @@ unions = List.foldl' union empty
 {-# INLINE unions #-}
 
 -- | /O(1)/ Return 'True' if this set is empty, 'False' otherwise.
+--
+-- >>> HashSet.null HashSet.empty
+-- True
+-- >>> HashSet.null (HashSet.singleton 1)
+-- False
 null :: HashSet a -> Bool
 null = H.null . asMap
 {-# INLINE null #-}
 
 -- | /O(n)/ Return the number of elements in this set.
+--
+-- >>> HashSet.size HashSet.empty
+-- 0
+-- >>> HashSet.size (HashSet.fromList [1,2,3])
+-- 3
 size :: HashSet a -> Int
 size = H.size . asMap
 {-# INLINE size #-}
 
 -- | /O(log n)/ Return 'True' if the given value is present in this
 -- set, 'False' otherwise.
+--
+-- >>> HashSet.member 1 (Hashset.fromList [1,2,3])
+-- True
+-- >>> HashSet.member 1 (Hashset.fromList [4,5,6])
+-- False
 member :: (Eq a, Hashable a) => a -> HashSet a -> Bool
 member a s = case H.lookup a (asMap s) of
                Just _ -> True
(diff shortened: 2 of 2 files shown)


m-renaud:
@sjakobi wdyt about this compared to having separate external docs / tutorial modules?

Data/HashSet.hs
```diff
@@ -4,25 +4,68 @@
...
+insertion, lookups, and deletion. If you are storing sets of @Int@s consider
+using the @Data.IntSet@ from the
+<https://hackage.haskell.org/packages/containers containers> package.
+A 'HashSet' makes no guarantees as to the order of its elements.
+
+== Immutability
+
+@HashSet@s are /immutable/ which means that any update functions do not modify
+the set you passed in, they create a new set. In order to keep the changes
+you need to assign it to a new variable.
```
sjakobi, line 27: Is this really necessary? We do assume that the reader has used Haskell before, right?

Data/HashSet.hs
```diff
@@ -4,25 +4,68 @@
...
+let s2 = HashSet.'HashSet.delete' "a" s1
+
+-- Print the @HashSet@ @s1@, notice it still contains the deleted @"a"@.
+print s1
+> HashSet.'HashSet.fromList' ["a","b"]
+
+-- Print the modified @HashSet@, notice that it does not contain the deleted @"a"@.
+print s2
+> HashSet.'HashSet.fromList' ["b"]
+@
```
sjakobi, line 50: See my commenthttps://github.com/haskell-unordered-containers/unordered-containers/pull/249#discussion_r436263795 on the other PR regarding the notation.

I think this would also be a good place to very briefly demonstrate some fundamental operations like `member`, `insert` and `delete`.

Data/HashSet.hs
```diff
@@ -4,25 +4,68 @@
...
+
+__IMPORTANT:__ @HashSet@ relies on the @element@ type having instances of the
+@Eq@ and @Hashable@ typeclasses for its internal representation. These are
+already defined for builtin types, and if you are using your own data type you
+can use the <https://en.wikibooks.org/wiki/Haskell/Classes_and_types#Deriving deriving>
```
sjakobi, line 55: That's not a great resource. For example it contains this misleading bit:

> You can only use deriving with a limited set of built-in classes, which are described very briefly below:

Data/HashSet.hs
```diff
@@ -4,25 +4,68 @@
...
+mechanism.
+
+== Performance
+
+The implementation is based on /hash array mapped trie/.  A
+'HashSet' is often faster than other tree-based set types,
+especially when value comparison is expensive, as in the case of
+strings.
```
sjakobi, line 63: "value comparison" seemed somewhat vague to me…

```suggestion
The implementation is based on /hash array mapped trie/.  A
'HashSet' is often faster than other 'Ord'-based set types,
especially when comparisons are expensive, as in the case of
strings.
```

Data/HashSet.hs
sjakobi: Maybe the best way to clarify this would be to have a short example that constructs a `HashSet` of a custom data type with derived instances.

Data/HashSet.hs
m-renaud: Done.

Data/HashSet.hs
m-renaud: Good idea.

Data/HashSet.hs
m-renaud: Done, PTAL!

Data/HashSet.hs
m-renaud: Eh, a lot of the time when you're first getting into a languages you'll reach for data structure libraries to play around simple problems, so from my perspective it doesn't hurt and is a concrete example for newcomers. I could go either way though.

m-renaud:
@sjakobi made some changes to formatting and added some comments as you suggested.

> What's the motivation for shuffling around the complexity annotations though. Doesn't that make the documentation much longer?

The complexity of the function isn't _the_ most important thing about the docs, and it feels kinda weird for the docs which describe what the functions do to start with _O(something)_. This also leaves space for commentary on the complexity (see 'HashSet.size' docs) while keeping the formatting consistent. It does make the docs longer, but I don't think optimizing for doc length over organization is the right trade-off.

m-renaud:
Also, switch from **bold** to _emphasis_ for the complexity clause on the functions at the bottom of the docs. I personally prefer the emphasis instead of bold because it doesn't compete with the function names for attention.

sjakobi:
> > What's the motivation for shuffling around the complexity annotations though. Doesn't that make the documentation much longer?
> 
> The complexity of the function isn't _the_ most important thing about the docs, and it feels kinda weird for the docs which describe what the functions do to start with _O(something)_. This also leaves space for commentary on the complexity (see 'HashSet.size' docs) while keeping the formatting consistent. It does make the docs longer, but I don't think optimizing for doc length over organization is the right trade-off.

I haven't really made up my mind about it, but I'd prefer to discuss it separately, i.e. not as a part of this PR.

> Also, switch from **bold** to _emphasis_ for the complexity clause on the functions at the bottom of the docs. I personally prefer the emphasis instead of bold because it doesn't compete with the function names for attention.

Which bold complexity clauses does this refer to?

m-renaud:
> I haven't really made up my mind about it, but I'd prefer to discuss it separately, i.e. not as a part of this PR.

Sure, I can move them back to being inline.

> Which bold complexity clauses does this refer to?

**_Complexity:_** _O(n)_

vs.

_Complexity:_ _O(n)_

Data/HashSet.hs
```diff
@@ -4,25 +4,99 @@
...
+To create a @HashSet@ of your custom type, you must derive @Eq@ and
+@Hashable@. The @Hashable@ typeclass is defined in the
+<https://hackage.haskell.org/packages/hashable hashable> package and the
+recommended way to define the instance is using generics, for which you'll need
+the @DeriveGeneric@ GHC extension.
+
+We'll start by setting up our custom data type:
+
+>>> :set -XDeriveGeneric
+>>> -- or {-# DeriveGeneric #-} if writing a Haskell source file.
```
sjakobi, line 76: This doesn't get rendered right, possibly due a GHC or Haddock bug:

image

This might work, but I didn't check:

```suggestion
>>> -- or \{-# DeriveGeneric #-} if writing a Haskell source file.
```

Data/HashSet.hs
```diff
@@ -4,25 +4,99 @@
...
+>>> import GHC.Generics (Generic)
+>>> import Data.Hashable
+>>> data Person = Person { name :: String, likesDogs :: Bool } deriving (Show, Eq, Generic)
+>>> instance Hashable Person
+
+And now we'll use it!
+
+>>> let people = HashSet.fromList [Person "Lana" True, Person "Joe" False, Person "Simon" True]
+>>> HashSet.filter likesDogs people
+fromList [Person {name = "Simon", likesDogs = True},Person {name = "Lana", likesDogs = True}]
```
sjakobi, line 86:
```
Person {name = "Simon", likesDogs = True}
```

How did you know?! ;)

Data/HashSet.hs
sjakobi: Yeah. I'm slightly worried about losing the reader with obvious things like that. So I'm still slightly in favour of removing this.

Data/HashSet.hs
```diff
@@ -4,25 +4,99 @@
...
+
+
+== Performance
+
+The implementation is based on /hash array mapped trie/.  A
```
sjakobi, line 91:
```suggestion
The implementation is based on /hash array mapped tries/.  A
```

?

Data/HashSet/Base.hs
```diff
@@ -259,38 +259,63 @@ fromListConstr = mkConstr hashSetDataType "fromList" [] Prefix
 hashSetDataType :: DataType
 hashSetDataType = mkDataType "Data.HashSet.Base.HashSet" [fromListConstr]
 
--- | /O(1)/ Construct an empty set.
+-- | Construct an empty set.
+--
+-- >>> HashSet.empty
```
sjakobi, line 264: I think the qualification is unnecessary for these examples.

```suggestion
-- >>> empty
```

Data/HashSet.hs
```diff
@@ -4,25 +4,99 @@
...
+Create a new set and combine it with our original set.
+
+>>> let unorderedDataStructures = HashSet.fromList ["HashSet", "HashMap"]
+>>> HashSet.union dataStructures unorderedDataStructures
+fromList ["Map","HashSet","Graph","HashMap","Set","Sequence"]
+
+=== Using custom data with HashSet
+
+To create a @HashSet@ of your custom type, you must derive @Eq@ and
+@Hashable@. The @Hashable@ typeclass is defined in the
```
sjakobi, line 68:
```suggestion
To create a @HashSet@ of your custom type, the type must have instances for
'Eq' and 'Hashable'. The 'Hashable' typeclass is defined in the
```

Data/HashSet.hs
m-renaud: No amount of `\`s inline made it render properly, so added a note above.

Data/HashSet.hs
m-renaud: :upside_down_face:

Data/HashSet.hs
m-renaud: Done.

Data/HashSet/Base.hs
m-renaud: I left it qualified because the container types should always be imported qualified because their names conflict, and `empty` also conflicts with the function from Control.Applicative. It would also be strange to use qualified functions everywhere else except for here.

sjakobi: Fair enough.

Data/HashSet.hs
```diff
@@ -4,25 +4,95 @@
...
+Stability   :  provisional
+Portability :  portable
+
+= Introduction
+
+'HashSet' allows you to store /unique/ elements, providing efficient
+insertion, lookups, and deletion. If you are storing sets of @Int@s consider
+using the @Data.IntSet@ from the
+<https://hackage.haskell.org/packages/containers containers> package.
+A 'HashSet' makes no guarantees as to the order of its elements.
```
sjakobi, line 21: It just seemed a bit strange to recommend an alternative data structure in the second sentence…

(Also, the URL was wrong)

```suggestion
'HashSet' allows you to store /unique/ elements, providing efficient
insertion, lookups, and deletion. A 'HashSet' makes no guarantees
as to the order of its elements.

If you are storing sets of @Int@s consider using "Data.IntSet" from the
<https://hackage.haskell.org/package/containers containers> package.
```

Data/HashSet.hs
```diff
@@ -4,25 +4,95 @@
...
+
+>>> let unorderedDataStructures = HashSet.fromList ["HashSet", "HashMap"]
+>>> HashSet.union dataStructures unorderedDataStructures
+fromList ["Map","HashSet","Graph","HashMap","Set","Sequence"]
+
+=== Using custom data with HashSet
+
+To create a @HashSet@ of your custom type, the type must have instances for 'Eq' and
+'Hashable'. The 'Hashable' typeclass is defined in the
+<https://hackage.haskell.org/packages/hashable hashable> package and the
```
sjakobi, line 63:
```suggestion
<https://hackage.haskell.org/package/hashable hashable> package and the
```

Data/HashSet.hs
```diff
@@ -4,25 +4,95 @@
...
+recommended way to define the instance is using generics, for which you'll need
+the @DeriveGeneric@ GHC extension.
+
+Note: You'll need to enable the @DeviceGeneric@ language extension, either via
+@:set -XDeriveGeneric@ in GHCi or @\{\-\# DeriveGeneric \#\-\}@ in your Haskell source
+code.
```
sjakobi, line 69: I think it's slightly unfortunate to go into so much detail here when the canonical documentation on this is in `hashable`: http://hackage.haskell.org/package/hashable-1.3.0.0/docs/Data-Hashable.html#g:3

Maybe just add a link?!

treeowl: No need to add anything. If @m-renaud wants that in the ReadTheDocs stuff, that's his prerogative, but it doesn't belong in the Haddocks.

sjakobi: @treeowl I think the current idea is not to add a ReadTheDocs page, but instead to include some examples and a bit of introductory documentation in the main haddocks.

I think there's some value in including an example of using u-c with custom datatypes, but I mostly feel that it shouldn't duplicate documentation that is or should be in hashable instead…

m-renaud: Yeah, my rationale for including a bit more than a link out was to prevent documentation hops, especially given that u-c is likely one of the first packages that someone new to the language may encounter. It does add a _bit_ more content to the docs but having a full working example in one place without having to piece things together can be very useful.

Additionally, the canonical documentation doesn't mention what to do if you're in ghci, so I thought including it here would be helpful as well for those not familiar with extensions. I don't think we should assume that anyone reading these docs knows that `:set -XExtensionName` is how to enable extensions in ghci; yes they're bound to figure it out eventually the more time you spend in the Haskell ecosystem, but one extra sentence to avoid a stackoverflow or reddit post seems like a good trade-off to me.

@sjakobi @treeowl please either :+1:  or :-1: this comment:

- :+1: - rationale sounds good, lets leave this
- :-1: - still don't agree, please remove details

Data/HashSet.hs
m-renaud: Good call.

Data/HashSet.hs
m-renaud: Fixed.

treeowl:
I'm a bit torn. I agree with much of your rationale, but I don't we want the u-c Haddocks to become an "introduction to Haskell". Would it make sense to put a full example with very little explanation in the module intro? Maybe. I'm sure we don't want to get into "how to enable extensions in GHCi". That's just far out of scope here.

m-renaud:
@treeowl how's this?

m-renaud:
@sjakobi should I do the HashMap docs in this PR as well or a follow-up?

sjakobi:
> @sjakobi should I do the HashMap docs in this PR as well or a follow-up?

Let's make a decision on `HashSet` first.

Data/HashSet.hs
```diff
@@ -4,25 +4,94 @@
...
+Create a new set and combine it with our original set.
+
+>>> let unorderedDataStructures = HashSet.fromList ["HashSet", "HashMap"]
+>>> HashSet.union dataStructures unorderedDataStructures
+fromList ["Map","HashSet","Graph","HashMap","Set","Sequence"]
+
+=== Using custom data with HashSet
+
+To create a @HashSet@ of your custom type, the type must have instances for 'Eq'
+and 'Hashable'. The 'Hashable' typeclass is defined in the
```
sjakobi, line 64: Haddock doesn't seem to be able to link `Eq` and `Hashable` since they both aren't in scope:

```
Warning: 'Eq' is out of scope.
    If you qualify the identifier, haddock can try to link it anyway.
Warning: 'Hashable' is out of scope.
    If you qualify the identifier, haddock can try to link it anyway.
```

The reason is this very limited import section below:

```
import Data.HashSet.Base
import Prelude ()
```

This would be good to fix, either by qualifying the identifiers or by modifying the imports.

Please the check the haddock warnings afterwards.

Data/HashSet.hs
sjakobi: I'm very happy with the section in its current state.

> u-c is likely one of the first packages that someone new to the language may encounter

This is surprising to me! Why do you think new users will encounter it so soon? u-c always seemed fairly niche compared to `containers` to me.

Data/HashSet.hs
```diff
@@ -4,25 +4,94 @@
...
+Stability   :  provisional
+Portability :  portable
+
+= Introduction
+
+'HashSet' allows you to store /unique/ elements, providing efficient insertion,
+lookups, and deletion. A 'HashSet' makes no guarantees as to the order of its
+elements.
+
+If you are storing sets of @Int@s consider using the @Data.IntSet@ from the
```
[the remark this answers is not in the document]
m-renaud: Done.

Data/HashSet.hs
```diff
@@ -4,25 +4,94 @@
...
+>>> -- Check if "Map" and "Trie" are in the set of data structures.
+>>> HashSet.member "Map" dataStructures
+True
+>>> HashSet.member "Trie" dataStructures
+False
+
+Add a new entry to the set:
+
+>>> let moreDataStructures = HashSet.insert "Trie" dataStructures
+>>> HashSet.member \"Trie\" moreDataStructures
```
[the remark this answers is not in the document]
m-renaud: Done.

Data/HashSet.hs
m-renaud: Opted for qualifying the identifiers. Changing the imports just to reference them in Haddocks then triggers unused import warnings (since ghc can't tell that the imports are only there for Haddocks).

m-renaud:
Done. ptal!

Data/HashSet.hs
sjakobi: Thanks!

I think you're at least in some instances linking to modules now instead of the types or type classes themselves. For example `"Data.Eq"`. I'd suggest linking to `'Data.Eq.Eq'` instead.

sjakobi:
Nearly done, IMHO! Great job!

Data/HashSet.hs
m-renaud: Updated here and for Data.Ord.Ord.

m-renaud:
Sorry for the delay, updated typeclass references. If there's nothing else I can squash and merge.

m-renaud:
jk, linknig to "Data.Eq.Eq" isn't valid :/

sjakobi:
> jk, linknig to "Data.Eq.Eq" isn't valid :/

It should be `'Data.Eq.Eq'` I think, with single quotes or alternatively backticks. I'll take another look tomorrow.

sjakobi:
> > jk, linknig to "Data.Eq.Eq" isn't valid :/
> 
> It should be `'Data.Eq.Eq'` I think, with single quotes or alternatively backticks. I'll take another look tomorrow.

Note that haddock might still produce a warning since the identifier is not in scope. But it should be able to produce the correct link anyway.

Also, please in include `Hashable`: `'Data.Hashable.Hashable'`.

sjakobi:
@treeowl Could you please review?

@m-renaud If you need help with the final touches to the markup, I can prepare a PR against yours.

sjakobi:
> @treeowl Could you please review?

Any @emilypi, too!? :)

emilypi:
I'll have time this weekend @sjakobi 😄

m-renaud:
I'll have time this weekend. Eq and Ord linked properly but Hashable didn't for some reason :/

m-renaud:
@sjakobi at least locally I can't get Hashable to link, any suggestions?

sjakobi:
> @sjakobi at least locally I can't get Hashable to link, any suggestions?

It actually works for me when I build the documentation with `cabal haddock --enable-documentation`.

What command did you use? I suspect you maybe just didn't build the haddocks for `hashable` too.

m-renaud:
For sure, thanks for the thorough review! I'll squash the commits and wait for the others to review before merging :)

sjakobi:
@treeowl, @emilypi I'd hate to see this PR languish. Could you please review?

If there are no reviews, I'll merge this on Monday (2020-07-20) or soon after.

sjakobi:
Thanks again, @m-renaud! :)

m-renaud:
No problem! Should I do the same for HashMap next?

sjakobi:
> Should I do the same for HashMap next?

IMHO, yes please! :)

I'd feel more comfortable if @treeowl and @emilypi would signal their approval too though.

PR #268, opened 2020-06-08
Remove dependency cycle hack from benchmark suite

andrewthad:


andrewthad:
After this, building the benchmarks can be done with:

    cabal build pkg:unordered-containers:benchmarks

And it no longer needlessly rebuilds all of the internal modules in `unordered-containers`.

PR #282, opened 2020-07-12
Fast inclusion operation on hashmaps and hashsets

svenkeidel:
Dear Johan, Dear David,

First of all, thanks for putting all this hard work into this library. We use it heavily in [Sturdy](https://github.com/svenkeidel/sturdy) to implement static code analyses in Haskell. To this end, we need a fast subset operation on hashmaps and hashsets. Unfortunately, while a naive implementation is easy to understand, it is rather slow because it compares keys with different hashes:
```Haskell
subset :: (Eq k, Hashable k, Eq v) => HashMap k v -> HashMap k v -> Bool
subset m1 m2 = keys m1 ⊆ keys m2 && and [ v1 == v2 | (k1,v1) <- toList m1; let v2 = m2 ! k1 ]
```

Instead, I worked on an [implementation](https://github.com/haskell-unordered-containers/unordered-containers/compare/master...svenkeidel:subset2?expand=1#diff-11321ff356efb0247b49763554dc8620R1431) that only compares keys with the same hash. I had to extend the `lookupCont` function with the number of the subkey to avoid code duplication. I tested this implementation with a few QuickCheck [properties](https://github.com/haskell-unordered-containers/unordered-containers/compare/master...svenkeidel:subset2?expand=1#diff-3d7e2d423eb3691821ef78fb73b2fa75R481-R489). What do you think?

I would be delighted if you would add this function to the library.

Regards,
Sven

### DIFF
--- a/Data/HashMap/Internal.hs
+++ b/Data/HashMap/Internal.hs
@@ -150,7 +150,7 @@ import qualified Data.Foldable as Foldable
 import Data.Bifoldable
 #endif
 import qualified Data.List as L
-import GHC.Exts ((==#), build, reallyUnsafePtrEquality#)
+import GHC.Exts ((==#), build, reallyUnsafePtrEquality#, inline)
 import Prelude hiding (filter, foldl, foldr, lookup, map, null, pred)
 import Text.Read hiding (step)
 
@@ -1431,8 +1431,8 @@ alterFEager f !k m = (<$> f mv) $ \fres ->
 -- >>> fromList [(1,'a'),(2,'b')] `isSubmapOf` fromList [(1,'a')]
 -- False
 isSubmapOf :: (Eq k, Hashable k, Eq v) => HashMap k v -> HashMap k v -> Bool
-isSubmapOf = isSubmapOfBy (==)
-{-# INLINE isSubmapOf #-}
+isSubmapOf = (inline isSubmapOfBy) (==)
+{-# INLINABLE isSubmapOf #-}
 
 -- | /O(n*log m)/ Inclusion of maps with value comparison. A map is included in
 -- another map if the keys are subsets and if the comparison function is true
@@ -1502,6 +1502,7 @@ isSubmapOfBy comp !m1 !m2 = go 0 m1 m2
     go _ (BitmapIndexed {}) (Collision {}) = False
     go _ (Full {}) (Collision {}) = False
     go _ (Full {}) (BitmapIndexed {}) = False
+{-# INLINABLE isSubmapOfBy #-}
 
 -- | /O(min n m))/ Checks if a bitmap indexed node is a submap of another.
 submapBitmapIndexed :: (HashMap k v1 -> HashMap k v2 -> Bool) -> Bitmap -> A.Array (HashMap k v1) -> Bitmap -> A.Array (HashMap k v2) -> Bool
@@ -1525,6 +1526,7 @@ submapBitmapIndexed comp !b1 !ary1 !b2 !ary2 = subsetBitmaps && go 0 0 (b1Orb2 .
     b1Andb2 = b1 .&. b2
     b1Orb2  = b1 .|. b2
     subsetBitmaps = b1Orb2 == b2
+{-# INLINABLE submapBitmapIndexed #-}
 
 ------------------------------------------------------------------------
 -- * Combine
--- a/Data/HashMap/Internal.hs
+++ b/Data/HashMap/Internal.hs
@@ -669,6 +669,10 @@ lookupRecordCollision h k m = lookupCont (\_ -> Absent) Present h k 0 m
 -- so we can be representation-polymorphic in the result type. Since
 -- this whole thing is always inlined, we don't have to worry about
 -- any extra CPS overhead.
+--
+-- The @Int@ argument is the offset of the subkey in the hash. When looking up
+-- keys at the top-level of a hashmap, the offset should be 0. When looking up
+-- keys at level @n@ of a hashmap, the offset should be @n * bitsPerSubkey@.
 lookupCont ::
 #if __GLASGOW_HASKELL__ >= 802
   forall rep (r :: TYPE rep) k v.
@@ -680,7 +684,7 @@ lookupCont ::
   -> (v -> Int -> r) -- Present continuation
   -> Hash -- The hash of the key
   -> k
-  -> Int -- The subkey
+  -> Int -- The offset of the subkey in the hash.
   -> HashMap k v -> r
 lookupCont absent present !h0 !k0 !s0 !m0 = go h0 k0 s0 m0
   where
--- a/Data/HashMap/Internal.hs
+++ b/Data/HashMap/Internal.hs
@@ -1417,7 +1417,7 @@ alterFEager f !k m = (<$> f mv) $ \fres ->
 {-# INLINABLE alterFEager #-}
 #endif
 
--- | /O(n*m)/ Inclusion of maps. A map is included in another map if the keys
+-- | /O(n*log m)/ Inclusion of maps. A map is included in another map if the keys
 -- are subsets and the corresponding values are equal:
 --
 -- >>> isSubmapOf m1 m2 = keys m1 ⊆ keys m2 && and [ v1 == v2 | (k1,v1) <- toList m1; let v2 = m2 ! k1 ]
@@ -1439,7 +1439,7 @@ isSubmapOf :: (Eq k, Hashable k, Eq v) => HashMap k v -> HashMap k v -> Bool
 isSubmapOf = isSubmapOfBy (==)
 {-# INLINE isSubmapOf #-}
 
--- | /O(n*m)/ Inclusion of maps with value comparison. A map is included in
+-- | /O(n*log m)/ Inclusion of maps with value comparison. A map is included in
 -- another map if the keys are subsets and the corresponding values are smaller:
 --
 -- >>> isSubmapOfBy (⊑) m1 m2 = keys m1 ⊆ keys m2 && and [ v1 ⊑ v2 | (k1,v1) <- toList m1; let v2 = m2 ! k1 ]
@@ -1450,16 +1450,14 @@ isSubmapOf = isSubmapOfBy (==)
 -- holds @isSubmapOfBy (⊑) m1 (unionWith (⊔) m1 m2)@ and
 -- @isSubmapOfBy (⊑) m2 (unionWith (⊔) m1 m2)@.
 isSubmapOfBy :: (Eq k, Hashable k) => (v1 -> v2 -> Bool) -> HashMap k v1 -> HashMap k v2 -> Bool
--- O(n*m) is the worst case runtime complexity. The worst case is when both
--- hashmaps m1 and m2 are collision nodes. Since collision nodes are unsorted
--- arrays, it requires for every key in m1 a linear search to to find a matching
--- key in m2, hence O(n*m).
+-- For maps without collisions the complexity is O(n*log m), where n is the size
+-- of m1 and m the size of m2: the inclusion operation visits every leaf in m1 at least once.
+-- For each leaf in m1, it looks up the key in m2.
 --
--- For maps without collisions the complexity is O(n), where n is the size of
--- m1: the inclusion operation traverses both maps in synchrony. When it
--- traverses down an edge in m1, it selects the appropriate edge in m2 with the
--- same hash prefix. Furthermore, it only needs to visit each node in m1 and m2
--- once. Therefore, the complexity is O(n).
+-- The worst case complexity is O(n*m). The worst case is when both hashmaps m1
+-- and m2 are collision nodes. Since collision nodes are unsorted arrays, it
+-- requires for every key in m1 a linear search to to find a matching key in m2,
+-- hence O(n*m).
 isSubmapOfBy comp = go 0
   where
     -- An empty map is always a submap of any other map.
--- a/Data/HashMap/Internal.hs
+++ b/Data/HashMap/Internal.hs
@@ -1422,15 +1422,18 @@ alterFEager f !k m = (<$> f mv) $ \fres ->
 --
 -- >>> isSubmapOf m1 m2 = keys m1 ⊆ keys m2 && and [ v1 == v2 | (k1,v1) <- toList m1; let v2 = m2 ! k1 ]
 --
--- This defines a partial order on maps, for which 'union' is the least upper bound.
--- More specifically, @isSubmapOf m1 (union m1 m2)@ and @isSubmapOf m2 (union m1 m2)@.
+-- This defines a partial order on maps. However, 'union' is /not/ its least
+-- upper bound, because it is not cummutative. In particular, for all maps @m1@
+-- and @m2@ it holds @m1 `isSubmapOf` union m1 m2@, however, the other direction
+-- @m2 `isSubmapOf` union m1 m2@ may not hold because 'union' may discard values
+-- of m2.
 --
 -- ==== __Examples__
 --
--- >>> isSubmapOf (fromList [(1,'a')]) (fromList [(1,'a'),(2,'b')])
+-- >>> fromList [(1,'a')] `isSubmapOf` fromList [(1,'a'),(2,'b')]
 -- True
 --
--- >>> isSubmapOf (fromList [(1,'a'),(2,'b')]) (fromList [(1,'a')])
+-- >>> fromList [(1,'a'),(2,'b')] `isSubmapOf` fromList [(1,'a')]
 -- False
 isSubmapOf :: (Eq k, Hashable k, Eq v) => HashMap k v -> HashMap k v -> Bool
 isSubmapOf = isSubmapOfBy (==)
@@ -1441,8 +1444,11 @@ isSubmapOf = isSubmapOfBy (==)
 --
 -- >>> isSubmapOfBy (⊑) m1 m2 = keys m1 ⊆ keys m2 && and [ v1 ⊑ v2 | (k1,v1) <- toList m1; let v2 = m2 ! k1 ]
 --
--- This defines a partial order on maps, for which 'unionWith' is the least upper bound.
--- More specifically, @isSubmapOfBy (⊑) m1 (unionWith (⊔) m1 m2)@ and @isSubmapOfBy (⊑) m2 (unionWith (⊔) m1 m2)@.
+-- This defines a partial order on maps, for which 'unionWith' is the least
+-- upper bound. More specifically, let @(⊑)@ be a partial order on values for
+-- which @(⊔)@ is the least upper bound. Then for all maps @m1@ and @m2@ it
+-- holds @isSubmapOfBy (⊑) m1 (unionWith (⊔) m1 m2)@ and
+-- @isSubmapOfBy (⊑) m2 (unionWith (⊔) m1 m2)@.
 isSubmapOfBy :: (Eq k, Hashable k) => (v1 -> v2 -> Bool) -> HashMap k v1 -> HashMap k v2 -> Bool
 isSubmapOfBy comp = go 0
   where
--- a/Data/HashMap/Internal.hs
+++ b/Data/HashMap/Internal.hs
@@ -1450,6 +1450,16 @@ isSubmapOf = isSubmapOfBy (==)
 -- holds @isSubmapOfBy (⊑) m1 (unionWith (⊔) m1 m2)@ and
 -- @isSubmapOfBy (⊑) m2 (unionWith (⊔) m1 m2)@.
 isSubmapOfBy :: (Eq k, Hashable k) => (v1 -> v2 -> Bool) -> HashMap k v1 -> HashMap k v2 -> Bool
+-- O(n*m) is the worst case runtime complexity. The worst case is when both
+-- hashmaps m1 and m2 are collision nodes. Since collision nodes are unsorted
+-- arrays, it requires for every key in m1 a linear search to to find a matching
+-- key in m2, hence O(n*m).
+--
+-- For maps without collisions the complexity is O(n), where n is the size of
+-- m1: the inclusion operation traverses both maps in synchrony. When it
+-- traverses down an edge in m1, it selects the appropriate edge in m2 with the
+-- same hash prefix. Furthermore, it only needs to visit each node in m1 once.
+-- Therefore, the complexity is O(n).
 isSubmapOfBy comp = go 0
   where
     -- An empty map is always a submap of any other map.
--- a/Data/HashMap/Internal.hs
+++ b/Data/HashMap/Internal.hs
@@ -1465,6 +1465,9 @@ isSubmapOfBy comp = go 0
     -- An empty map is always a submap of any other map.
     go !_ Empty _ = True
 
+    -- If the second map is empty and the first is not, it cannot be a submap.
+    go _ _ Empty = False
+
     -- If the first map contains only one entry, lookup the key in the second map.
     go s (Leaf h1 (L k1 v1)) t2 = lookupCont (\_ -> False) (\v2 _ -> comp v1 v2) h1 k1 s t2
 
@@ -1500,12 +1503,6 @@ isSubmapOfBy comp = go 0
       submapBitmapIndexed (go (s+bitsPerSubkey)) fullNodeMask ls1 fullNodeMask ls2
     go s (Full ls1) (BitmapIndexed b2 ls2) =
       submapBitmapIndexed (go (s+bitsPerSubkey)) fullNodeMask ls1 b2 ls2
-
-    -- TODO: I'm not sure about these cases and need help. If we cleared up all
-    -- these cases, we can replace them with a catch-all case go _ _ _ = False
-
-    -- If the second map is empty and the first is not, it cannot be a submap.
-    go _ _ Empty = False
 
     -- A collision, bitmap indexed an full node always contain at least two
     -- entries. Hence it cannot be a map of a leaf.
--- a/Data/HashMap/Internal.hs
+++ b/Data/HashMap/Internal.hs
@@ -1457,10 +1457,10 @@ isSubmapOfBy :: (Eq k, Hashable k) => (v1 -> v2 -> Bool) -> HashMap k v1 -> Hash
 -- and m2 are collision nodes for the same hash. Since collision nodes are
 -- unsorted arrays, it requires for every key in m1 a linear search to to find a
 -- matching key in m2, hence O(n*m).
-isSubmapOfBy comp = go 0
+isSubmapOfBy comp !m1 !m2 = go 0 m1 m2
   where
     -- An empty map is always a submap of any other map.
-    go !_ Empty _ = True
+    go _ Empty _ = True
 
     -- If the second map is empty and the first is not, it cannot be a submap.
     go _ _ Empty = False
@@ -1505,10 +1505,10 @@ isSubmapOfBy comp = go 0
 
 -- | /O(min n m))/ Checks if a bitmap indexed node is a submap of another.
 submapBitmapIndexed :: (HashMap k v1 -> HashMap k v2 -> Bool) -> Bitmap -> A.Array (HashMap k v1) -> Bitmap -> A.Array (HashMap k v2) -> Bool
-submapBitmapIndexed comp b1 ary1 b2 ary2 = subsetBitmaps && go 0 0 (b1Orb2 .&. negate b1Orb2)
+submapBitmapIndexed comp !b1 !ary1 !b2 !ary2 = subsetBitmaps && go 0 0 (b1Orb2 .&. negate b1Orb2)
   where
     go :: Int -> Int -> Bitmap -> Bool
-    go i j m
+    go !i !j !m
       | m > b1Orb2 = True
 
       -- In case a key is both in ary1 and ary2, check ary1[i] <= ary2[j] and
--- a/Data/HashMap/Internal.hs
+++ b/Data/HashMap/Internal.hs
@@ -49,8 +49,8 @@ module Data.HashMap.Internal
     , update
     , alter
     , alterF
-    , subset
-    , subsetWith
+    , isSubmapOf
+    , isSubmapOfBy
 
       -- * Combine
       -- ** Union
@@ -1417,40 +1417,40 @@ alterFEager f !k m = (<$> f mv) $ \fres ->
 {-# INLINABLE alterFEager #-}
 #endif
 
--- | /O(n*m)/ Subset of maps. A map is a subset of another map if the keys are
--- subsets and the corresponding values are equal:
+-- | /O(n*m)/ Inclusion of maps. A map is included in another map if the keys
+-- are subsets and the corresponding values are equal:
 --
--- >>> subset m1 m2 = keys m1 ⊆ keys m2 && and [ v1 == v2 | (k1,v1) <- toList m1; let v2 = m2 ! k1 ]
+-- >>> isSubmapOf m1 m2 = keys m1 ⊆ keys m2 && and [ v1 == v2 | (k1,v1) <- toList m1; let v2 = m2 ! k1 ]
 --
 -- This defines a partial order on maps, for which 'union' is the least upper bound.
--- More specifically, @subset m1 (union m1 m2)@ and @subset m2 (union m1 m2)@.
+-- More specifically, @isSubmapOf m1 (union m1 m2)@ and @isSubmapOf m2 (union m1 m2)@.
 --
 -- ==== __Examples__
 --
--- >>> subset (fromList [(1,'a')]) (fromList [(1,'a'),(2,'b')])
+-- >>> isSubmapOf (fromList [(1,'a')]) (fromList [(1,'a'),(2,'b')])
 -- True
 --
--- >>> subset (fromList [(1,'a'),(2,'b')]) (fromList [(1,'a')])
+-- >>> isSubmapOf (fromList [(1,'a'),(2,'b')]) (fromList [(1,'a')])
 -- False
-subset :: (Eq k, Hashable k, Eq v) => HashMap k v -> HashMap k v -> Bool
-subset = subsetWith (const (==))
-{-# INLINE subset #-}
+isSubmapOf :: (Eq k, Hashable k, Eq v) => HashMap k v -> HashMap k v -> Bool
+isSubmapOf = isSubmapOfBy (==)
+{-# INLINE isSubmapOf #-}
 
--- | /O(n*m)/ Subset of maps with value comparison. A map is a subset of another
--- map if the keys are subsets and the corresponding values are smaller:
+-- | /O(n*m)/ Inclusion of maps with value comparison. A map is included in
+-- another map if the keys are subsets and the corresponding values are smaller:
 --
--- >>> subsetWith (⊑) m1 m2 = keys m1 ⊆ keys m2 && and [ v1 ⊑ v2 | (k1,v1) <- toList m1; let v2 = m2 ! k1 ]
+-- >>> isSubmapOfBy (⊑) m1 m2 = keys m1 ⊆ keys m2 && and [ v1 ⊑ v2 | (k1,v1) <- toList m1; let v2 = m2 ! k1 ]
 --
 -- This defines a partial order on maps, for which 'unionWith' is the least upper bound.
--- More specifically, @subsetWith (⊑) m1 (unionWith (⊔) m1 m2)@ and @subset (⊑) m2 (unionWith (⊔) m1 m2)@.
-subsetWith :: (Eq k, Hashable k) => (k -> v1 -> v2 -> Bool) -> HashMap k v1 -> HashMap k v2 -> Bool
-subsetWith comp = go 0
+-- More specifically, @isSubmapOfBy (⊑) m1 (unionWith (⊔) m1 m2)@ and @isSubmapOfBy (⊑) m2 (unionWith (⊔) m1 m2)@.
+isSubmapOfBy :: (Eq k, Hashable k) => (v1 -> v2 -> Bool) -> HashMap k v1 -> HashMap k v2 -> Bool
+isSubmapOfBy comp = go 0
   where
-    -- An empty map is always a subset of any other map.
+    -- An empty map is always a submap of any other map.
     go !_ Empty _ = True
 
     -- If the first map contains only one entry, lookup the key in the second map.
-    go s (Leaf h1 (L k1 v1)) t2 = lookupCont (\_ -> False) (\v2 _ -> comp k1 v1 v2) h1 k1 s t2
+    go s (Leaf h1 (L k1 v1)) t2 = lookupCont (\_ -> False) (\v2 _ -> comp v1 v2) h1 k1 s t2
 
     -- In this case we need to check that for each x in ls1, there is a y in ls2
     -- such that x ⊑ y. This is the worst case complexity-wise since it requires a O(m*n) check.
@@ -1477,33 +1477,30 @@ subsetWith comp = go 0
     -- In cases where the first and second map are bitmap indexed or full,
     -- traverse down the tree at the appropriate indices.
     go s (BitmapIndexed b1 ls1) (BitmapIndexed b2 ls2) =
-      subsetBitmapIndexed (go (s+bitsPerSubkey)) b1 ls1 b2 ls2
+      submapBitmapIndexed (go (s+bitsPerSubkey)) b1 ls1 b2 ls2
     go s (BitmapIndexed b1 ls1) (Full ls2) =
-      subsetBitmapIndexed (go (s+bitsPerSubkey)) b1 ls1 fullNodeMask ls2
+      submapBitmapIndexed (go (s+bitsPerSubkey)) b1 ls1 fullNodeMask ls2
     go s (Full ls1) (Full ls2) =
-      subsetBitmapIndexed (go (s+bitsPerSubkey)) fullNodeMask ls1 fullNodeMask ls2
+      submapBitmapIndexed (go (s+bitsPerSubkey)) fullNodeMask ls1 fullNodeMask ls2
     go s (Full ls1) (BitmapIndexed b2 ls2) =
-      subsetBitmapIndexed (go (s+bitsPerSubkey)) fullNodeMask ls1 b2 ls2
+      submapBitmapIndexed (go (s+bitsPerSubkey)) fullNodeMask ls1 b2 ls2
 
     -- TODO: I'm not sure about these cases and need help. If we cleared up all
     -- these cases, we can replace them with a catch-all case go _ _ _ = False
 
-    -- If the second map is empty, but the first is not, it cannot be a subset.
+    -- If the second map is empty and the first is not, it cannot be a submap.
     go _ _ Empty = False
 
-    -- A collision always contains at least two entries. Hence it cannot be a
-    -- subset of a leaf.
+    -- A collision, bitmap indexed an full node always contain at least two
+    -- entries. Hence it cannot be a map of a leaf.
     go _ (Collision {}) (Leaf {}) = False
-
-    -- A bitmap indexed node and a full node always contain at least two
-    -- entries. Hence they cannot be a subset of a leaf.
     go _ (BitmapIndexed {}) (Leaf {}) = False
     go _ (Full {}) (Leaf {}) = False
 
 
--- | /O(min n m))/ hecks if a bitmap indexed node is a subset of another.
-subsetBitmapIndexed :: (HashMap k v1 -> HashMap k v2 -> Bool) -> Bitmap -> A.Array (HashMap k v1) -> Bitmap -> A.Array (HashMap k v2) -> Bool
-subsetBitmapIndexed comp b1 ary1 b2 ary2 = subsetBitmaps && go 0 0 (b1Orb2 .&. negate b1Orb2)
+-- | /O(min n m))/ Checks if a bitmap indexed node is a submap of another.
+submapBitmapIndexed :: (HashMap k v1 -> HashMap k v2 -> Bool) -> Bitmap -> A.Array (HashMap k v1) -> Bitmap -> A.Array (HashMap k v2) -> Bool
+submapBitmapIndexed comp b1 ary1 b2 ary2 = subsetBitmaps && go 0 0 (b1Orb2 .&. negate b1Orb2)
   where
     go :: Int -> Int -> Bitmap -> Bool
     go i j m
@@ -2191,13 +2188,13 @@ updateOrConcatWithKey f ary1 ary2 = A.run $ do
 {-# INLINABLE updateOrConcatWithKey #-}
 
 -- | /O(n*m)/ Check if the first array is a subset of the second array.
-subsetArray :: Eq k => (k -> v1 -> v2 -> Bool) -> A.Array (Leaf k v1) -> A.Array (Leaf k v2) -> Bool
+subsetArray :: Eq k => (v1 -> v2 -> Bool) -> A.Array (Leaf k v1) -> A.Array (Leaf k v2) -> Bool
 subsetArray cmpV ary1 ary2 = A.length ary1 <= A.length ary2 && go 0 (A.length ary1)
   where
     go !i n
       | i >= n = True
       | otherwise = let (L k1 v1) = A.index ary1 i
-                    in lookupInArrayCont (\_ -> False) (\v2 _ -> cmpV k1 v1 v2 && go (i+1) n) k1 ary2
+                    in lookupInArrayCont (\_ -> False) (\v2 _ -> cmpV v1 v2 && go (i+1) n) k1 ary2
 
 ------------------------------------------------------------------------
 -- Manually unrolled loops
--- a/Data/HashMap/Internal.hs
+++ b/Data/HashMap/Internal.hs
@@ -1476,7 +1476,7 @@ isSubmapOfBy comp = go 0
     go _ (Collision h1 ls1) (Collision h2 ls2) =
       h1 == h2 && subsetArray comp ls1 ls2
 
-    -- To check ls1 ⊆ ls2, we only need to check the entries in ls2 with the hash h1.
+    -- To check t1 ⊆ t2, we only need to check the entries in ls2 with the hash h1.
     go s t1@(Collision h1 _) (BitmapIndexed b ls2)
         | b .&. m == 0 = False
         | otherwise    =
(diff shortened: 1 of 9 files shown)


treeowl:
Thanks for all this work! Can you add an explanation of what you mean by a "subkey"?

Data/HashMap/Base.hs
```diff
@@ -1398,6 +1402,112 @@ alterFEager f !k m = (<$> f mv) $ \fres ->
...
+-- More specifically, @subset m1 (union m1 m2)@ and @subset m2 (union m1 m2)@.
+--
+-- ==== __Examples__
+--
+-- >>> subset (fromList [(1,'a')]) (fromList [(1,'a'),(2,'b')])
+-- True
+--
+-- >>> subset (fromList [(1,'a'),(2,'b')]) (fromList [(1,'a')])
+-- False
+subset :: (Eq k, Hashable k, Eq v) => HashMap k v -> HashMap k v -> Bool
```
sjakobi, line 1424: In `containers`, this function is named [`isSubmapOf`](http://hackage.haskell.org/package/containers-0.6.2.1/docs/Data-Map-Strict-Internal.html#v:isSubmapOf).

Data/HashMap/Base.hs
```diff
@@ -1398,6 +1406,112 @@ alterFEager f !k m = (<$> f mv) $ \fres ->
...
+{-# INLINE subset #-}
+
+-- | /O(n*m)/ Subset of maps with value comparison. A map is a subset of another
+-- map if the keys are subsets and the corresponding values are smaller:
+--
+-- >>> subsetWith (⊑) m1 m2 = keys m1 ⊆ keys m2 && and [ v1 ⊑ v2 | (k1,v1) <- toList m1; let v2 = m2 ! k1 ]
+--
+-- This defines a partial order on maps, for which 'unionWith' is the least upper bound.
+-- More specifically, @subsetWith (⊑) m1 (unionWith (⊔) m1 m2)@ and @subset (⊑) m2 (unionWith (⊔) m1 m2)@.
+subsetWith :: (Eq k, Hashable k) => (k -> v1 -> v2 -> Bool) -> HashMap k v1 -> HashMap k v2 -> Bool
```
sjakobi, line 1435: The corresponding function in `containers` is named [`isSubmapOfBy`](http://hackage.haskell.org/package/containers-0.6.2.1/docs/Data-Map-Strict-Internal.html#v:isSubmapOfBy), and it its comparison function doesn't have access to the keys. Is the access to the keys important?

treeowl:
The not-fun part: how does this change affect benchmarks? Also also, do you have ideas for implementing #226? That would presumably subsume this.

Data/HashSet/Base.hs
```diff
@@ -282,6 +283,10 @@ fromMap = HashSet
 keysSet :: HashMap k a -> HashSet k
 keysSet m = fromMap (() <$ m)
 
+-- | /O(n*m)/ Subset of sets.
+subset :: (Eq a, Hashable a) => HashSet a -> HashSet a -> Bool
```
sjakobi, line 287: The `containers` variant is named [`isSubsetOf`](http://hackage.haskell.org/package/containers-0.6.2.1/docs/Data-Set.html#v:isSubsetOf).

svenkeidel:
> treeowl: how does this change affect benchmarks?

You mean the performance of `lookupCont` (the only existing function I changed) and functions that use `lookupCont`? I have not tested this yet. The generated code should be the same though, since `lookupCont` is inlined and its recursive helper function `go` already uses a subkey offset.

> treeowl: do you have ideas for implementing #226? That would presumably subsume this.

You mean something like `subset = mergeA (const False) (const True) (==)`? I guess this would work. I don't know if this implementation would be as efficient as what I implemented right now. I don't know how to implement this. I presume an implementation of `merge` is closer to `union` than to `subset`.

> sjakobi: I have a few comments regarding the consistency with the containers API.

Thanks, I will change these function names.

svenkeidel:
One thing that is missing is a proper benchmark for this operation. Any ideas for what would be a good benchmark?

Data/HashMap/Base.hs
svenkeidel: No, I guess key access is not important. I'll remove it.

svenkeidel:
@treeowl, I compared the performance of `subset2` vs `master` and calculated the speedup (slowdown): [performance.xlsx](https://github.com/haskell-unordered-containers/unordered-containers/files/4911199/performance.xlsx)

The speedup is within 0.97 and 1.04, which I would attribute to measuring inaccuracies.

svenkeidel:
@sjakobi, I renamed the functions for compatibility with the containers API in https://github.com/svenkeidel/unordered-containers/commit/ffa499e7c8a087a44442992af9d266c412e962c0. Github did not show this commit here because of an outage.

sjakobi:
> @sjakobi, I renamed the functions for compatibility with the containers API in [svenkeidel@ffa499e](https://github.com/svenkeidel/unordered-containers/commit/ffa499e7c8a087a44442992af9d266c412e962c0). Github did not show this commit here because of an outage.

Could you try pushing it again? The outage is supposed to be over: https://www.githubstatus.com/

Data/HashMap/Base.hs
```diff
@@ -1398,6 +1406,115 @@ alterFEager f !k m = (<$> f mv) $ \fres ->
...
+-- | /O(n*m)/ Inclusion of maps with value comparison. A map is included in
+-- another map if the keys are subsets and the corresponding values are smaller:
+--
+-- >>> isSubmapOfBy (⊑) m1 m2 = keys m1 ⊆ keys m2 && and [ v1 ⊑ v2 | (k1,v1) <- toList m1; let v2 = m2 ! k1 ]
+--
+-- This defines a partial order on maps, for which 'unionWith' is the least
+-- upper bound. More specifically, let @(⊑)@ be a partial order on values for
+-- which @(⊔)@ is the least upper bound. Then for all maps @m1@ and @m2@ it
+-- holds @isSubmapOfBy (⊑) m1 (unionWith (⊔) m1 m2)@ and
+-- @isSubmapOfBy (⊑) m2 (unionWith (⊔) m1 m2)@.
```
sjakobi, line 1440: I'm not a huge fan of using `⊑` as the name of the comparison function.

1. I can't directly type it on my keyboard.
2. It looks very similar to `⊆`

So I'd slightly prefer a name like `cmpV` or similar.

Data/HashMap/Base.hs
```diff
@@ -1398,6 +1406,115 @@ alterFEager f !k m = (<$> f mv) $ \fres ->
...
+
+    -- If the first map contains only one entry, lookup the key in the second map.
+    go s (Leaf h1 (L k1 v1)) t2 = lookupCont (\_ -> False) (\v2 _ -> comp v1 v2) h1 k1 s t2
+
+    -- In this case we need to check that for each x in ls1, there is a y in ls2
+    -- such that x ⊑ y. This is the worst case complexity-wise since it requires a O(m*n) check.
+    go _ (Collision h1 ls1) (Collision h2 ls2) =
+      h1 == h2 && subsetArray comp ls1 ls2
+
+    -- To check ls1 ⊆ ls2, we only need to check the entries in ls2 with the hash h1.
```
sjakobi, line 1455: ls1 is not "in scope".

Data/HashMap/Base.hs
```diff
@@ -1398,6 +1406,115 @@ alterFEager f !k m = (<$> f mv) $ \fres ->
...
+    go s t1@(Collision h1 _) (Full ls2) =
+      go (s+bitsPerSubkey) t1 (A.index ls2 (index h1 s))
+
+    -- Insert the collision into a bitmap indexed node and recurse.
+    go s t1@(BitmapIndexed {}) t2@(Collision h2 _) =
+      go s t1 (BitmapIndexed (mask h2 s) (A.singleton t2))
+    go s t1@(Full {}) t2@(Collision h2 _) =
+      go s t1 (BitmapIndexed (mask h2 s) (A.singleton t2))
+
+    -- In cases where the first and second map are bitmap indexed or full,
```
sjakobi, line 1472:
```suggestion
    -- In cases where the first and second map are BitmapIndexed or Full,
```

Data/HashMap/Base.hs
```diff
@@ -1398,6 +1406,115 @@ alterFEager f !k m = (<$> f mv) $ \fres ->
...
+      submapBitmapIndexed (go (s+bitsPerSubkey)) b1 ls1 b2 ls2
+    go s (BitmapIndexed b1 ls1) (Full ls2) =
+      submapBitmapIndexed (go (s+bitsPerSubkey)) b1 ls1 fullNodeMask ls2
+    go s (Full ls1) (Full ls2) =
+      submapBitmapIndexed (go (s+bitsPerSubkey)) fullNodeMask ls1 fullNodeMask ls2
+    go s (Full ls1) (BitmapIndexed b2 ls2) =
+      submapBitmapIndexed (go (s+bitsPerSubkey)) fullNodeMask ls1 b2 ls2
+
+    -- TODO: I'm not sure about these cases and need help. If we cleared up all
+    -- these cases, we can replace them with a catch-all case go _ _ _ = False
```
sjakobi, line 1484: Catch-alls like that IMHO rather hurt readability and make it less obvious that the code is correct. GHC can often simplify to produce such catch-alls anyway.

Data/HashMap/Base.hs
```diff
@@ -1398,6 +1406,115 @@ alterFEager f !k m = (<$> f mv) $ \fres ->
...
+
+    -- If the second map is empty and the first is not, it cannot be a submap.
+    go _ _ Empty = False
```
sjakobi, line 1487: Maybe move this directly below `go !_ Empty _ = True`?

Data/HashMap/Base.hs
```diff
@@ -1398,6 +1406,115 @@ alterFEager f !k m = (<$> f mv) $ \fres ->
...
+
+    -- A collision, bitmap indexed an full node always contain at least two
```
sjakobi, line 1489: Where is this documented? (And is it checked?)

Data/HashSet/Base.hs
```diff
@@ -282,6 +283,10 @@ fromMap = HashSet
 keysSet :: HashMap k a -> HashSet k
 keysSet m = fromMap (() <$ m)
 
+-- | /O(n*m)/ Inclusion of sets.
```
sjakobi, line 286: An example would be nice to have here.

Data/HashMap/Base.hs
```diff
@@ -1398,6 +1406,115 @@ alterFEager f !k m = (<$> f mv) $ \fres ->
 {-# INLINABLE alterFEager #-}
 #endif
 
+-- | /O(n*m)/ Inclusion of maps. A map is included in another map if the keys
```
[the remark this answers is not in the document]
svenkeidel: `O(n*m)` is the _worst case_ runtime complexity. The worst case is when both hashmaps `m1` and `m2` are collision nodes (https://github.com/svenkeidel/unordered-containers/blob/subset2/Data/HashMap/Base.hs#L1450-L1453). Since collision nodes are unsorted arrays, it requires for every key in `m1` a linear search to to find a matching key in `m2`, hence `O(n*m)`. 

That being said, the worst case is not very likely though. We hope that with a good choice of hashing function collisions are not very likely. For maps without collisions the complexity is more like `O(n)`, where `n` is the size of `m1`: By exploiting the structure of array mapped hash tries, you only need to look at one node in `m2` for every node in `m1`, to find a matching pairs of keys.

By the way, this shows that only documenting the worst-case complexity is not very useful if we think that it is very unlikely. We should also document the complexity for cases where the hashmaps have no collisions.

Data/HashMap/Base.hs
svenkeidel: I used this symbol, because it is well established in static analyses and order theory (e.g. see https://cs.au.dk/~amoeller/spa/spa.pdf sec 4.2). They use `⊑` as an ordering symbol and `⊔` as least upper bounds symbol for a lattice `(⊑,⊔)`. Furthermore, the similarity of `⊑` to `⊆` is not by accident: Powersets form a lattice with `(⊆,∪)`. Lastly, you don't need to be able to type these symbols on your keyboard, because they only appear in this documentation and not in the code of users.

Data/HashMap/Base.hs
svenkeidel: Yes, these are the cases I'm not sure about, hence the TODO above.

The implementation assumes that this is an invariant of hashmaps. Therefore it is not checked.

Data/HashMap/Base.hs
svenkeidel: I would prefer to keep this notation for clarity sake.

tests/HashMapProperties.hs
```diff
@@ -225,6 +225,45 @@ pAlterFLookup k f =
   `eq`
   getConst . HM.alterF (Const . apply f) k
 
+pSubmapReflexive :: [(Key, Int)] -> Bool
+pSubmapReflexive xs =
+  let m = HM.fromList xs
```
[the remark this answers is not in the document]
svenkeidel: I tried to switch to `HashMap Key Int`, however, there is no `Arbitrary` instance yet. Should I create one for now until #154 is resolved?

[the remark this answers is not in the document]
sjakobi: Oh, apparently the instance is in `tests/Strictness.hs`. Maybe it could be shared? I'm not very familiar with the structure of the testsuites yet.

Duplicating it here seems good enough too, though.

Or leave it for future work. It's not really your responsibility IMHO.

Data/HashMap/Base.hs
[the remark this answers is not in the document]
sjakobi: AFAICT most other complexity annotations refer to the "average case". I think it would be good to stay consistent with this pattern.

I think it would be good to start documenting the worst case complexity too though. Maybe include it further down in the haddocks?

Data/HashMap/Base.hs
sjakobi: > I used this symbol, because it is well established in static analyses and order theory (e.g. see https://cs.au.dk/~amoeller/spa/spa.pdf sec 4.2). They use `⊑` as an ordering symbol and `⊔` as least upper bounds symbol for a lattice `(⊑,⊔)`. Furthermore, the similarity of `⊑` to `⊆` is not by accident: Powersets form a lattice with `(⊆,∪)`. Lastly, you don't need to be able to type these symbols on your keyboard, because they only appear in this documentation and not in the code of users.

I appreciate the background info! :)

I don't have a strong math background and I don't want to expect users to have it. I think it's important that the documentation is accessible to users from a wide range of backgrounds.

I'm also slightly worried that these mathematical symbols might cause an instinctive reaction like "Oh, it's math – I'm not smart enough for that".

But maybe I'm the outlier here with this view.

@m-renaud, @emilypi What do you think?

Data/HashMap/Base.hs
sjakobi: I'm somewhat concerned that e.g. the results of `delete` or `difference` might violate these assumptions, but I don't know for sure.

@treeowl Do you know more about this?

@svenkeidel I think this is related: https://github.com/haskell-unordered-containers/unordered-containers/issues/161#issuecomment-311246147

Data/HashMap/Base.hs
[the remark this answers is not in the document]
svenkeidel: Ok, after thinking about it, the average runtime of `isSubmapOf m1 m2` is `O(n*log m)`, where `n = size m1` and `m = size m2`.
Proof: `isSubmapOf` visits every leaf in m1 at least once. For each leaf in m1, it looks up its key in m2.

treeowl:
I think mathematical symbols are fine, but I like them presented as MathJax, not Unicode in the source.

Data/HashMap/Base.hs
svenkeidel: Ok, I removed them from the documentation.

svenkeidel:
@sjakobi, @treeowl thanks for the feedback. Up to this last issue (https://github.com/haskell-unordered-containers/unordered-containers/pull/282#discussion_r453606996), I addressed all feedback.

Data/HashMap/Base.hs
svenkeidel: This is a case which reduces the size of a `BitmapIndexed` node to 1: https://github.com/haskell-unordered-containers/unordered-containers/blob/master/Data/HashMap/Base.hs#L1096

I'll implement these cases instead of relying that this invariant holds.

treeowl: It's not inherently a problem to have a `BitmapIndexed` of length 1, is it? Or is it? Let's work this out for sure. I'd prefer not to have to change the `Eq` instance.

svenkeidel: It is not a problem. But I relied on this assumption in the implementation of `isSubmapOf`. In other words, it is a bug in `isSubmapOf` I still need to fix.

svenkeidel: But first I want to understand why my QuickCheck tests did not catch this bug.

sjakobi: @svenkeidel If you could create a reproducer that violates these assumptions, that would be very useful, and not just for this PR! :)

Data/HashSet/Base.hs
```diff
@@ -282,6 +283,18 @@ fromMap = HashSet
...
```
sjakobi, line 286:
```suggestion
-- | /O(n*log m)/ Inclusion of sets.
```

?

Data/HashMap/Base.hs
```diff
@@ -1398,6 +1406,119 @@ alterFEager f !k m = (<$> f mv) $ \fres ->
...
+-- True
+--
+-- >>> fromList [(1,'a'),(2,'b')] `isSubmapOf` fromList [(1,'a')]
+-- False
+isSubmapOf :: (Eq k, Hashable k, Eq v) => HashMap k v -> HashMap k v -> Bool
+isSubmapOf = isSubmapOfBy (==)
+{-# INLINE isSubmapOf #-}
+
+-- | /O(n*log m)/ Inclusion of maps with value comparison. A map is included in
+-- another map if the keys are subsets and the corresponding values are smaller:
```
sjakobi, line 1427: "smaller" seems a bit confusing to me. How about something like this:

```suggestion
-- another map if the keys are subsets and if the comparison function is true for the corresponding values:
```

Data/HashMap/Base.hs
```diff
@@ -1398,6 +1406,119 @@ alterFEager f !k m = (<$> f mv) $ \fres ->
...
+--
+-- >>> isSubmapOfBy (<=) (fromList [(1,'b')]) (fromList [(1,'a'),(2,'c')])
+-- False
+isSubmapOfBy :: (Eq k, Hashable k) => (v1 -> v2 -> Bool) -> HashMap k v1 -> HashMap k v2 -> Bool
+-- For maps without collisions the complexity is O(n*log m), where n is the size
+-- of m1 and m the size of m2: the inclusion operation visits every leaf in m1 at least once.
+-- For each leaf in m1, it looks up the key in m2.
+--
+-- The worst case complexity is O(n*m). The worst case is when both hashmaps m1
+-- and m2 are collision nodes. Since collision nodes are unsorted arrays, it
```
sjakobi, line 1445:
```suggestion
-- and m2 are collision nodes for the same hash. Since collision nodes are unsorted arrays, it
```

?

Data/HashMap/Base.hs
```diff
@@ -2065,6 +2186,15 @@ updateOrConcatWithKey f ary1 ary2 = A.run $ do
...
 {-# INLINABLE updateOrConcatWithKey #-}
 
+-- | /O(n*m)/ Check if the first array is a subset of the second array.
+subsetArray :: Eq k => (v1 -> v2 -> Bool) -> A.Array (Leaf k v1) -> A.Array (Leaf k v2) -> Bool
+subsetArray cmpV ary1 ary2 = A.length ary1 <= A.length ary2 && go 0 (A.length ary1)
+  where
+    go !i n
+      | i >= n = True
+      | otherwise = let (L k1 v1) = A.index ary1 i
+                    in lookupInArrayCont (\_ -> False) (\v2 _ -> cmpV v1 v2 && go (i+1) n) k1 ary2
```
sjakobi, line 2196: The `lookupInArrayCont`-expression is quite a mouthful! ;)

Is this complexity "essential"? Or could this be implemented more readably, for example with a `foldr` over `ary1`?

I think we could also exploit the fact that once we have found a corresponding element at a certain index in `ary2`, we don't have to check that index again. Does that sound worthwhile?

Data/HashMap/Base.hs
treeowl: Ah, I see. The invalid assumption you made is that the array in a `BitmapIndexed` node has at least two elements. The invariant that *does* hold (or had better hold, anyway) is that when a `BitMapIndexed` array only has one element, that element is *not* a `Leaf` or a `Collision`. This, along with the ban on internal `Empty` nodes, ensures that map representation is unique up to collision bucket permutation.

svenkeidel: This is very helpful. We should document this in the code, best at the type definition.

One more question. From studying `delete`, I learned that `Full` nodes are downgraded to `BitmapIndexed` nodes as soon as a child node becomes empty. Similarly, `Collision` nodes are downgraded to `Leafs` when there is only a single key left. I assume this means that `Full` and `Collision` nodes have at least 2 elements. Correct?

treeowl: `Collision` nodes have at least two. `Full` nodes have *exactly* sixteen, where 16 comes about from some fixed definition in the module. (Don't remember what it's called, or exactly how it gives 16.)

Data/HashMap/Base.hs
svenkeidel: I changed the implementation to use `Array.all`. It looks a bit more readable now.

svenkeidel:
Thanks again for your feedback. I addressed all further comments. This is about all time I can and want to spend on this PR.

Data/HashMap/Base.hs
```diff
@@ -1398,6 +1406,121 @@ alterFEager f !k m = (<$> f mv) $ \fres ->
...
+
+      -- In case a key occurs in ary1, but not ary2, only increment index j.
+      | b2 .&. m /= 0 = go i (j+1) (m `unsafeShiftL` 1)
+
+      -- In case a key neither occurs in ary1 nor ary2, continue.
+      | otherwise = go i j (m `unsafeShiftL` 1)
+
+    b1Andb2 = b1 .&. b2
+    b1Orb2  = b1 .|. b2
+    subsetBitmaps = b1Orb2 == b2
```
sjakobi, line 1523: This bit goes way over my head. Can I rely on you, @treeowl, to review this?

Data/HashMap/Base.hs
```diff
@@ -2065,6 +2188,13 @@ updateOrConcatWithKey f ary1 ary2 = A.run $ do
     return mary
 {-# INLINABLE updateOrConcatWithKey #-}
 
+-- | /O(n*m)/ Check if the first array is a subset of the second array.
+subsetArray :: Eq k => (v1 -> v2 -> Bool) -> A.Array (Leaf k v1) -> A.Array (Leaf k v2) -> Bool
+subsetArray cmpV ary1 ary2 = A.length ary1 <= A.length ary2 && A.all inAry2 ary1
+  where
+    inAry2 (L k1 v1) = lookupInArrayCont (\_ -> False) (\v2 _ -> cmpV v1 v2) k1 ary2
+    {-# INLINE inAry2 #-}
```
sjakobi, line 2196: @svenkeidel, @treeowl What did you think about my idea to record the indices of any found elements in `ary2` and to skip these when checking the remaining elements of `ary1`.

I'm mostly wondering whether that's an idea worth recording – I wouldn't insist that it be included in this PR.

svenkeidel: It is hard to say if this is going to make a big difference performance wise. For each key you still would need to iterate over ary2 anyways. The only thing you would save are a few equality checks of keys. It could be worth working on this optimization, but before we need some benchmarks to verify that we actually improve the performance.

Data/HashMap/Base.hs
svenkeidel: If it helps, I copied some of this code from `unionArrayBy` (https://github.com/haskell-unordered-containers/unordered-containers/blob/7e8fc9ec5304b743a2f54b8c299c5ea1425080b7/Data/HashMap/Base.hs#L1642-L1657).

treeowl: > This bit goes way over my head. Can I rely on you, @treeowl, to review this?

Mine eyes glazeth over when it comes to bit fiddling. I can *try*. Mostly, I rely on testing.

svenkeidel:
@sjakobi, about benchmarks. My collaborator Tobias Hombücher is currently working on a static analysis for Scheme that uses this hashmaps and hashsets extensively (https://gitlab.rlp.net/plmz/sturdy/-/tree/PowersetT/scheme). We have a benchmark suit with a few scheme programs (https://gitlab.rlp.net/plmz/sturdy/-/blob/PowersetT/scheme/bench/FixpointBench.hs). Once Tobias gets the analysis working, this would be a great real-world benchmark for the unordered-containers library.

That being said, a micro benchmark for the `isSubmapOf` operation would still be valuable and I still want to implement it.

svenkeidel:
I added a few benchmarks. There is still some room for improvement.
```
benchmarked Map/isSubmapOf/String
time                 153.7 μs   (152.7 μs .. 155.5 μs)
mean                 154.0 μs   (153.6 μs .. 154.7 μs)
std dev              1.609 μs   (953.4 ns .. 2.458 μs)

benchmarked Map/isSubmapOf/ByteString
time                 80.00 μs   (78.90 μs .. 81.17 μs)
mean                 79.74 μs   (79.42 μs .. 80.96 μs)
std dev              1.673 μs   (632.2 ns .. 3.537 μs)

benchmarked hashmap/Map/isSubmapOf/String
time                 135.3 μs   (134.5 μs .. 136.2 μs)
mean                 136.4 μs   (135.9 μs .. 137.5 μs)
std dev              2.301 μs   (1.123 μs .. 3.720 μs)

benchmarked hashmap/Map/isSubmapOf/ByteString
time                 78.98 μs   (78.52 μs .. 79.67 μs)
                     1.000 R²   (0.999 R² .. 1.000 R²)
mean                 78.60 μs   (78.42 μs .. 78.92 μs)
std dev              703.4 ns   (428.3 ns .. 1.199 μs)

benchmarked IntMap/isSubmapOf
time                 128.5 ns   (126.6 ns .. 131.1 ns)
                     0.997 R²   (0.994 R² .. 1.000 R²)
mean                 128.8 ns   (128.1 ns .. 130.1 ns)
std dev              2.688 ns   (1.236 ns .. 4.315 ns)

benchmarked HashMap/isSubmapOf/String
time                 199.8 μs   (198.7 μs .. 200.7 μs)
mean                 202.5 μs   (201.5 μs .. 204.4 μs)
std dev              4.169 μs   (2.111 μs .. 6.863 μs)

benchmarked HashMap/isSubmapOf/ByteString
time                 101.0 μs   (100.0 μs .. 102.1 μs)
mean                 100.5 μs   (100.3 μs .. 101.2 μs)
std dev              1.285 μs   (683.3 ns .. 2.341 μs)

benchmarked HashMap/isSubmapOf/Int
time                 83.80 ns   (80.81 ns .. 86.37 ns)
                     0.963 R²   (0.884 R² .. 0.999 R²)
mean                 85.79 ns   (84.39 ns .. 90.96 ns)
std dev              6.445 ns   (1.616 ns .. 13.01 ns)
variance introduced by outliers: 36% (moderately inflated)

benchmarked HashMap/isSubmapOfNaive/String
time                 218.3 μs   (214.8 μs .. 222.4 μs)
mean                 221.3 μs   (220.1 μs .. 223.4 μs)
std dev              5.203 μs   (3.151 μs .. 8.042 μs)

benchmarked HashMap/isSubmapOfNaive/ByteString
time                 104.8 μs   (103.6 μs .. 105.8 μs)
                     0.992 R²   (0.970 R² .. 1.000 R²)
mean                 106.6 μs   (105.6 μs .. 110.1 μs)
std dev              5.373 μs   (967.1 ns .. 11.16 μs)
variance introduced by outliers: 28% (moderately inflated)

benchmarked HashMap/isSubmapOfNaive/Int
time                 84.29 ns   (82.98 ns .. 85.34 ns)
                     0.999 R²   (0.998 R² .. 0.999 R²)
mean                 83.93 ns   (83.48 ns .. 84.41 ns)
std dev              1.231 ns   (941.8 ps .. 1.615 ns)
```

Data/HashMap/Base.hs
```diff
@@ -1398,6 +1406,121 @@ alterFEager f !k m = (<$> f mv) $ \fres ->
...
+            go (s+bitsPerSubkey) t1 (A.index ls2 (sparseIndex b m))
+      where m = mask h1 s
+
+    -- Similar to the previous case we need to traverse l2 at the index for the hash h1.
+    go s t1@(Collision h1 _) (Full ls2) =
+      go (s+bitsPerSubkey) t1 (A.index ls2 (index h1 s))
+
+    -- Insert the collision into a bitmap indexed node and recurse.
+    go s t1@(BitmapIndexed {}) t2@(Collision h2 _) =
+      go s t1 (BitmapIndexed (mask h2 s) (A.singleton t2))
```
treeowl, line 1479: This seems weird.

Data/HashMap/Base.hs
```diff
@@ -1398,6 +1406,121 @@ alterFEager f !k m = (<$> f mv) $ \fres ->
...
+    go s t1@(BitmapIndexed {}) t2@(Leaf h2 _) =
+      go s t1 (BitmapIndexed (mask h2 s) (A.singleton t2))
```
treeowl, line 1481: Huh? I think you should just have a case up above for the second map being a `Leaf`. The first map is only a submap of a leaf if it's empty (already covered) or it's also a leaf and things compare right.

Data/HashMap/Base.hs
```diff
@@ -1398,6 +1406,121 @@ alterFEager f !k m = (<$> f mv) $ \fres ->
...
+    go s t1@(Full {}) t2@(Collision h2 _) =
+      go s t1 (BitmapIndexed (mask h2 s) (A.singleton t2))
+
```
treeowl, line 1484: Can you explain?

treeowl:
General question: would it be useful to implement a lookup-by-hash function that returns the `Leaf` or `Collision` node? That should help with `Collision` cases here, and would likely also be useful elsewhere.

Data/HashMap/Base.hs
sjakobi: @treeowl That's a neat idea! :)

If we don't do this in this PR, I'll record it in an issue.

Data/HashMap/Base.hs
```diff
@@ -1398,6 +1406,121 @@ alterFEager f !k m = (<$> f mv) $ \fres ->
...
+      submapBitmapIndexed (go (s+bitsPerSubkey)) b1 ls1 fullNodeMask ls2
+    go s (Full ls1) (Full ls2) =
+      submapBitmapIndexed (go (s+bitsPerSubkey)) fullNodeMask ls1 fullNodeMask ls2
+    go s (Full ls1) (BitmapIndexed b2 ls2) =
+      submapBitmapIndexed (go (s+bitsPerSubkey)) fullNodeMask ls1 b2 ls2
+
+    -- Collision and Full nodes always contain at least two entries. Hence it
+    -- cannot be a map of a leaf.
+    go _ (Collision {}) (Leaf {}) = False
+    go _ (Full {}) (Leaf {}) = False
```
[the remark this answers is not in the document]
svenkeidel: Isn't `isSubmapOfBy` a bit large for `INLINE`? This would create a lot of code bloat for every call of `isSubmapOfBy`. Why not `INLINABLE`? Then users can still specialize it within their code base.

Data/HashMap/Base.hs
```diff
@@ -1398,6 +1406,121 @@ alterFEager f !k m = (<$> f mv) $ \fres ->
...
+-- ==== __Examples__
+--
+-- >>> fromList [(1,'a')] `isSubmapOf` fromList [(1,'a'),(2,'b')]
+-- True
+--
+-- >>> fromList [(1,'a'),(2,'b')] `isSubmapOf` fromList [(1,'a')]
+-- False
+isSubmapOf :: (Eq k, Hashable k, Eq v) => HashMap k v -> HashMap k v -> Bool
+isSubmapOf = isSubmapOfBy (==)
+{-# INLINE isSubmapOf #-}
```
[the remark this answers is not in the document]
treeowl: We never need this to be inlined at a call site. But it is very valuable to specialize to particular key and value types. Once those types are known, I believe we want GHC to create a specialization that inlines the comparison function into `isSubmapOfBy`.

Data/HashMap/Base.hs
svenkeidel: I thought that through the process of deletion, the `BitmapIndexed` node `t1` could contain a single collision, which may be equal to the collision `t2`.

But I think you have a point. Based on our discussion a `BitmapIndexed` node is a map that contains more than 1 leaf/collision node. 

Therefore, this case should be
```Haskell
go s (BitmapIndexed {}) (Collision {}) = False
```

Correct?

Data/HashMap/Base.hs
svenkeidel: Yes, this should probably also be
```Haskell
go s (BitmapIndexed {}) (Leaf {}) = False
```

Data/HashMap/Base.hs
```diff
@@ -1398,6 +1406,121 @@ alterFEager f !k m = (<$> f mv) $ \fres ->
...
+
+    -- In cases where the first and second map are BitmapIndexed or Full,
+    -- traverse down the tree at the appropriate indices.
+    go s (BitmapIndexed b1 ls1) (BitmapIndexed b2 ls2) =
+      submapBitmapIndexed (go (s+bitsPerSubkey)) b1 ls1 b2 ls2
+    go s (BitmapIndexed b1 ls1) (Full ls2) =
+      submapBitmapIndexed (go (s+bitsPerSubkey)) b1 ls1 fullNodeMask ls2
+    go s (Full ls1) (Full ls2) =
+      submapBitmapIndexed (go (s+bitsPerSubkey)) fullNodeMask ls1 fullNodeMask ls2
+    go s (Full ls1) (BitmapIndexed b2 ls2) =
```
svenkeidel, line 1493: This case should also be `False`.

Data/HashMap/Base.hs
treeowl: Yes, I believe so. Because the `BitmapIndexed` or `Full` will always contain at least two hashes.

Data/HashMap/Base.hs
treeowl: Yup! This is simplifying nicely.

svenkeidel:
> General question: would it be useful to implement a lookup-by-hash function that returns the Leaf or Collision node? That should help with Collision cases here, and would likely also be useful elsewhere.

Yes, this function would allow us to remove a bit of code duplication. Furthermore, we can probably implement the regular `lookup` function with a lookup-by-hash function.

treeowl:
How do the benchmarks look with the simplifications in place?

svenkeidel:
The simplifications only changed negative cases where `isSubmapOf m1 m2 == False`. Therefore, the performance of the benchmarks did not change, since they are a positive test.

Data/HashMap/Base.hs
```diff
@@ -1398,6 +1406,114 @@ alterFEager f !k m = (<$> f mv) $ \fres ->
...
+    go _ (Collision {}) (Leaf {}) = False
+    go _ (BitmapIndexed {}) (Leaf {}) = False
+    go _ (Full {}) (Leaf {}) = False
+    go _ (BitmapIndexed {}) (Collision {}) = False
+    go _ (Full {}) (Collision {}) = False
+    go _ (Full {}) (BitmapIndexed {}) = False
+
+-- | /O(min n m))/ Checks if a bitmap indexed node is a submap of another.
+submapBitmapIndexed :: (HashMap k v1 -> HashMap k v2 -> Bool) -> Bitmap -> A.Array (HashMap k v1) -> Bitmap -> A.Array (HashMap k v2) -> Bool
+submapBitmapIndexed comp b1 ary1 b2 ary2 = subsetBitmaps && go 0 0 (b1Orb2 .&. negate b1Orb2)
```
treeowl, line 1497: Please bang everything but the function. As written, this is lazy in the arrays. I don't know if that leads to reboxing, but better safe than sorry.

Data/HashMap/Base.hs
```diff
@@ -1398,6 +1406,114 @@ alterFEager f !k m = (<$> f mv) $ \fres ->
...
+  where
+    go :: Int -> Int -> Bitmap -> Bool
+    go i j m
```
treeowl, line 1500: Bang all these too, just in case.

Data/HashMap/Base.hs
```diff
@@ -1398,6 +1406,114 @@ alterFEager f !k m = (<$> f mv) $ \fres ->
...
+-- For each leaf in m1, it looks up the key in m2.
+--
+-- The worst case complexity is O(n*m). The worst case is when both hashmaps m1
+-- and m2 are collision nodes for the same hash. Since collision nodes are
+-- unsorted arrays, it requires for every key in m1 a linear search to to find a
+-- matching key in m2, hence O(n*m).
+isSubmapOfBy comp = go 0
+  where
+    -- An empty map is always a submap of any other map.
+    go !_ Empty _ = True
```
treeowl, line 1452: Do we need/want to be lazy in the second argument here? If we really want that, we should probably do it at the top, not in the `go` function.

svenkeidel:
Something is fishy with the benchmarks. When I run only a single benchmark with `stack bench --ba='--match=prefix HashMap/isSubmapOf/String'`, I get
```
benchmarked HashMap/isSubmapOf/String
time                 167.6 μs   (166.8 μs .. 169.1 μs)
mean                 168.4 μs   (167.9 μs .. 169.4 μs)
std dev              2.105 μs   (1.232 μs .. 4.253 μs)
```
When I run multiple benchmarks with `stack bench --ba='--match=pattern isSubmapOf'`, I get
```
benchmarked HashMap/isSubmapOf/String
time                 205.8 μs   (204.8 μs .. 207.4 μs)
mean                 206.6 μs   (206.0 μs .. 207.8 μs)
std dev              2.561 μs   (1.487 μs .. 4.375 μs)
```
Either garbage collection kicks in, or there is something wrong with the benchmark environment `Env`. Eitherway, it should not make a difference when you run a benchmark alone or in a group.

treeowl:
Definitely an issue. Could caching play a role too? I don't know enough about these things.

sjakobi:
@svenkeidel You might be running into https://github.com/bos/criterion/issues/166?! 

You can try the `--measure-with` option: https://github.com/vincenthz/hs-gauge/pull/3

svenkeidel:
@sjakobi, `--measure-with` fixed it. Nonetheless, I think it is a bad idea to share a single benchmark environment across all benchmarks. It is reallocated for every benchmark group. We should split the environment into smaller environments for each benchmark group to reduce memory pressure.

sjakobi:
> Nonetheless, I think it is a bad idea to share a single benchmark environment across all benchmarks. It is reallocated for every benchmark group. We should split the environment into smaller environments for each benchmark group to reduce memory pressure.

I don't have a good understanding of how `env` is supposed to be used. Feel free to change it!

Data/HashMap/Base.hs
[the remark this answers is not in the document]
treeowl: I'm not sure. We want to be sure to inline it into `isSubmapOf`; does that happen if this is just marked `INLINABLE` (and that is too)?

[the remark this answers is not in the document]
treeowl: We can collapse these three `Leaf` cases into one and put it up above, matching the approach we take to `Empty`:

```haskell
    go s (Leaf h1 (L k1 v1)) t2 = lookupCont (\_ -> False) (\v2 _ -> comp v1 v2) h1 k1 s t2
    go _ _ (Leaf {}) = False
```

svenkeidel:
I started to refactor the benchmark suite. I mainly reduced code duplication and improved the reproducibility of the benchmark suite by reducing the memory allocations. I found quite a few bugs due to code duplication (e.g. [this](https://github.com/haskell-unordered-containers/unordered-containers/blob/master/benchmarks/Benchmarks.hs#L123-L125) should use `elemsDup`).

Data/HashMap/Base.hs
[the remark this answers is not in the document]
svenkeidel: > I'm not sure. We want to be sure to inline it into isSubmapOf; does that happen if this is just marked INLINABLE (and that is too)?

Another option is to mark both functions INLINABLE and then inline the `isSubmapOfBy` function at the `isSubmapOf` call site with the [`inline`](https://hackage.haskell.org/package/base-4.14.0.0/docs/GHC-Exts.html#v:inline) function.

benchmarks/Benchmarks.hs
```diff
@@ -25,305 +26,258 @@ import qualified Util.String as US
...
+    benchHashMap f = bgroup "HashMap" $
+      [ env (setupString HM.fromList)     $ \e -> bgroup "String" $ f e
+      , env (setupByteString HM.fromList) $ \e -> bgroup "ByteString" $ f e
+      , env (setupInt HM.fromList)        $ \e -> bgroup "Int" $ f e
+      ]
+  #-}
+{-# RULES
+  "force-inline-benchIntMap" forall (f :: Env IntMap Int Int -> [Benchmark]).
+    benchIntMap f = env (setupInt (IntMap . IM.fromList)) $ \e -> bgroup "IntMap" $ f e
+  #-}
```
sjakobi, line 228: I feel somewhat suspicious about adding this kind of complexity to the benchmarks.

Could you say why this is needed, @svenkeidel?

I also wonder whether this PR is the best place to take a deep dive into the benchmarking code. To simplify the review process, I'd slightly prefer keeping this about the original feature request, and to do any larger work on the benchmarks in a separate PR.

svenkeidel: > I feel somewhat suspicious about adding this kind of complexity to the benchmarks. Could you say why this is needed, @svenkeidel?

As I explained above, the refactoring 1. deduplicates the benchmarks and 2. reduces the number of allocations do make the benchmarks more reproducible. The rewrite rules ensure that benchmarking code is inlined, such that GHC can specialized it. I explain this in this comment: https://github.com/haskell-unordered-containers/unordered-containers/pull/282/commits/57acaec0434b8dffcc8debb9d985cad96e2239a2#diff-e00f41037166e0f4ab0362d1218e6926R179-R181
If we would use INLINE pragmas instead of rewrite rules, GHC may not inline and specialize the benchmark code because the inlined code is too large. Rewrite rules are the best option for forcing the inlining.

> I also wonder whether this PR is the best place to take a deep dive into the benchmarking code. To simplify the review process, I'd slightly prefer keeping this about the original feature request, and to do any larger work on the benchmarks in a separate PR.

I know that this is not the right place to do this. The problem is, as is, the benchmarks are unreliable and not reproducible (different runs produce different results). Therefore, it is hard for me to optimize the performance of `isSubmapOf` without reliable benchmarks.

We can remove the changes to the benchmark suite before we merge this PR. But for now I want to debug the performance with the current benchmark suite.

svenkeidel: Maybe I put the changes to the benchmark suite in a different branch on top of `subset2`.

sjakobi: > If we would use INLINE pragmas instead of rewrite rules, GHC may not inline and specialize the benchmark code because the inlined code is too large. Rewrite rules are the best option for forcing the inlining.

What do you mean with "may not"? Can't you tell from the Core?

I just generally feel that long-running feature branches are a bit of an anti-pattern, since they naturally accumulate merge conflicts and often end up stalling. My impression was that we were largely happy with the library changes, so I'd suggest merging them first, and doing further performance tweaks and/or benchmark refactorings in follow-up PRs.

svenkeidel: Agreed.

svenkeidel:
Ok, I rebased everything on master, removed the changes to the benchmark suite and squashed everything into one commit on this branch: https://github.com/svenkeidel/unordered-containers/tree/subset-squashed

sjakobi:
@svenkeidel Is it OK if we finish the review in this PR, so the commit to `master` is associated with the discussion, or do you want to open a separate PR from that branch?

svenkeidel:
The former makes most sense to me.

Data/HashMap/Internal.hs
```diff
@@ -1409,6 +1417,116 @@ alterFEager f !k m = (<$> f mv) $ \fres ->
...
+--
+-- ==== __Examples__
+--
+-- >>> fromList [(1,'a')] `isSubmapOf` fromList [(1,'a'),(2,'b')]
+-- True
+--
+-- >>> fromList [(1,'a'),(2,'b')] `isSubmapOf` fromList [(1,'a')]
+-- False
+isSubmapOf :: (Eq k, Hashable k, Eq v) => HashMap k v -> HashMap k v -> Bool
+isSubmapOf = (inline isSubmapOfBy) (==)
```
sjakobi, line 1434: AFAIK this is the first use of `inline` in this library. Why is it necessary?

Data/HashMap/Internal.hs
```diff
@@ -1409,6 +1417,116 @@ alterFEager f !k m = (<$> f mv) $ \fres ->
...
+
+    -- Collision and Full nodes always contain at least two entries. Hence it
+    -- cannot be a map of a leaf.
+    go _ (Collision {}) (Leaf {}) = False
+    go _ (BitmapIndexed {}) (Leaf {}) = False
+    go _ (Full {}) (Leaf {}) = False
+    go _ (BitmapIndexed {}) (Collision {}) = False
+    go _ (Full {}) (Collision {}) = False
+    go _ (Full {}) (BitmapIndexed {}) = False
+{-# INLINABLE isSubmapOfBy #-}
```
sjakobi, line 1505: IIRC @treeowl suggested

```suggestion
{-# INLINE isSubmapOfBy #-}
```

Or is there a reason for preferring `INLINABLE`?

Data/HashSet/Internal.hs
```diff
@@ -310,6 +311,18 @@ fromMap = HashSet
...
+--
+-- ==== __Examples__
+--
+-- >>> fromList [1,3] `isSubsetOf` fromList [1,2,3]
+-- True
+--
+-- >>> fromList [1,2] `isSubsetOf` fromList [1,3]
+-- False
+isSubsetOf :: (Eq a, Hashable a) => HashSet a -> HashSet a -> Bool
+isSubsetOf s1 s2 = H.isSubmapOfBy (\_ _ -> True) (asMap s1) (asMap s2)
```
sjakobi, line 324:
```suggestion
isSubsetOf s1 s2 = H.isSubmapOfBy (\_ _ -> True) (asMap s1) (asMap s2)
{-# INLINABLE isSubsetOf #-}
```

Data/HashMap/Internal.hs
```diff
@@ -1409,6 +1417,116 @@ alterFEager f !k m = (<$> f mv) $ \fres ->
...
+      -- In case a key occurs in ary1, but not ary2, only increment index j.
+      | b2 .&. m /= 0 = go i (j+1) (m `unsafeShiftL` 1)
+
+      -- In case a key neither occurs in ary1 nor ary2, continue.
+      | otherwise = go i j (m `unsafeShiftL` 1)
+
+    b1Andb2 = b1 .&. b2
+    b1Orb2  = b1 .|. b2
+    subsetBitmaps = b1Orb2 == b2
+{-# INLINABLE submapBitmapIndexed #-}
```
sjakobi, line 1529: What's the reason for using `INLINABLE` instead of `INLINE` here?

I suspect we want the whole definition to be inlined into `isSubmapOfBy`?!

treeowl:
I haven't reviewed them specifically, but rewrite rules in the benchmark suite make me very nervous in principle, because that takes the suite further from realistic code.

svenkeidel:
@treeowl, don't worry about them, I removed them from the PR.

sjakobi:
Oh, if my questions about the pragmas cannot be answered with the current state of the benchmark suite, then never mind.

It's fine for me if this is addressed in follow-up patches.

Data/HashMap/Internal.hs
sjakobi: Actually functions like `differenceWith` or `intersectionWithKey` also use `INLINABLE`. So using it here seems consistent.

Data/HashMap/Internal.hs
svenkeidel: See https://github.com/haskell-unordered-containers/unordered-containers/pull/282#discussion_r458593216

Data/HashMap/Internal.hs
svenkeidel: In general, I would avoid marking functions as INLINE that are too large or that use their argument more than once, e.g. `foo x = (x,x)`. This blows up the size of the generated code unnecessarily.

As @treeowl mentioned, to improve performance it makes sense to specialize the library functions to specific key types. However, GHC "compiles away" all code without any INLINE/INLINABLE pragmas. This means GHC cannot specialize these functions anymore, because it does not have access to their code anymore. To solve this, you can mark functions as INLINABLE to ask GHC to keep the code around, but not necessarily inline it.

sjakobi: That makes sense to me. @treeowl, what do you think?

treeowl: Could be. Let's do it for now And try to remember to check later if it works properly.

treeowl: Could be. Let's do it for now And try to remember to check later if it works properly.

Data/HashSet/Internal.hs
sjakobi: If INLINABLE helps with specialization, it would be good to use here, right?

treeowl: The best thing to do is always to write some sample code and pore over the Core. But that's not a hard requirement when adding new functions.

sjakobi:
@treeowl Are you done with reviewing this PR?

treeowl:
I want to make sure someone's verified that this doesn't hurt `lookup` or `alterF` performance.

sjakobi:
> I want to make sure someone's verified that this doesn't hurt `lookup` or `alterF` performance.

@svenkeidel What did your benchmark spreadsheet say about this? I unfortunately can't find it in this thread anymore.

svenkeidel:
https://github.com/haskell-unordered-containers/unordered-containers/pull/282#issuecomment-657408910

treeowl:
Please squash to one or two logical commits.

svenkeidel:
@treeowl, see https://github.com/haskell-unordered-containers/unordered-containers/pull/282#issuecomment-662495559

treeowl:
I'll let you and @sjakobi work out the merge mechanics.

svenkeidel:
@treeowl, @sjakobi, thanks for all the reviewing work and merging this PR :+1:

sjakobi:
You're welcome, @svenkeidel!

And thanks for the feedback on the test and benchmark setup! :) Any further help with that would be very much appreciated.

Would you mind if I'd prepare a release now, or do you see anything that needs to be done before that?

svenkeidel:
> Would you mind if I'd prepare a release now, or do you see anything that needs to be done before that?

Not at all. Sounds good :+1:

PR #303, opened 2021-01-22
Fix 'compose' since tag

Mathnerd314:
On [Hackage](https://hackage.haskell.org/package/unordered-containers-0.2.13.0/docs/Data-HashMap-Strict.html#g:6) there is an ugly `@since UNRELEASED` in the documentation for compose. I guess #299 didn't fix the tag when it got merged.

### DIFF
--- a/Data/HashMap/Internal.hs
+++ b/Data/HashMap/Internal.hs
@@ -1699,7 +1699,7 @@ unions = L.foldl' union empty
 -- ('compose' bc ab '!?') = (bc '!?') <=< (ab '!?')
 -- @
 --
--- @since UNRELEASED
+-- @since 0.2.13.0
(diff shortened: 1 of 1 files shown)


Mathnerd314:
Also, there are both "Since 0.2.10.0" and "Since 0.2.11". I'm not sure which version number style is preferred... I went with the more precise 0.2.13.0 here.

sjakobi (approved):
Thank you! I usually update the `@since`-annotations while preparing a release. Must have forgotten about that with the last release.

I don't have a clear preference regarding the version style right now.

treeowl:
The versioning policy doesn't allow new functions in sub-minor releases,
does it? So I think three-part versions probably make more sense there.

treeowl:
Thanks.

PR #305, opened 2021-02-06
add more NFData instances

### LINKED ISSUE #304: `instance NFData k => NFData1 (HashMap k)` and `instance NFData2 HashMap`
I'm filing this issue to request NFData1 and NFData2 instances for HashMap. If no one else does, I'll write the instances in the next 48 hours and submit a pull request.

LightAndLight:
adds:
* NFData2 HashMap
* NFData k => NFData1 (HashMap k)
* NFData1 HashSet

closes #304


treeowl:
I'm all in favor of the new instances. Please make sure performance is not lost for existing instances. Either

1. Check that they produce identical Core (including unfoldings, arity, and unfolding guidance)
2. Benchmark with several instances, especially including cheap ones (e.g., ones where `deepseq = seq`, where `rnf = const ()`, and where `rnf` just forces one field of a record), where the map is various sizes and already in normal form.
3. Leave the existing instances alone and just add the new ones on top. This is a bit silly, of course, but it's certainly the easiest way to be confident you won't goof anything up.

LightAndLight:
Will inspection-testing be sufficient for proving (1)?

treeowl:
No, I don't think so. That's not looking at things like unfoldings.

LightAndLight:
Given that, I decided to leave the existing instances as-is and only add the new ones.

sjakobi:
Can you add some CPP, so the new instances are only defined with recent enough versions of `deepseq`?

treeowl:
Or should we bump the minimum deepseq bound?

sjakobi:
> Or should we bump the minimum deepseq bound?

I think that might limit GHC compatibility too much for people who can't easily upgrade boot libraries separately from the compiler.

treeowl:
The latest deepseq supports base>=4.5. Even we don't go back that far!

sjakobi:
> The latest deepseq supports base>=4.5. Even we don't go back that far!

True. But AFAIK using the latest `deepseq` with e.g. GHC 8.2 is rather tricky with e.g. Nix.

treeowl:
Explain?

sjakobi:
I don't remember the details. AFAIK it's non-trivial to update boot libraries with Nix. I think I once tried to update `containers` or `bytestring` and it was a PITA.

treeowl:
For this we'd only need deepseq-1.4.3, which came out in July, 2017.
Honestly, if someone has to fuss a little too use that with an old GHC, I
don't think that's our problem.

sjakobi:
OK. If that's easier we'll need to update the `deepseq` bound and fix CI.

treeowl:
It's easy either way, but CPP always makes the future harder.

treeowl:
CI is failing.

LightAndLight:
Yes, it is. Have you reached consensus on how you'd like me to fix the
version issue? It looks like it leans toward "increase the deepseq bound".

treeowl:
That's definitely my inclination, yes.

sjakobi:
> Have you reached consensus on how you'd like me to fix the version issue? It looks like it leans toward "increase the deepseq bound".

My preference is for retaining backward compatibility with `deepseq < 1.4.3`, but raising the bound doesn't seem too bad either.

sjakobi:
On third thought, I'm fine with raising the `deepseq` bound for now. If this ends up causing problems for users we can probably improve compatibility later on.

@LightAndLight Could you raise the bound and fix CI, please?

unordered-containers.cabal
```diff
@@ -197,7 +197,7 @@ benchmark benchmarks
     bytestring,
     containers,
     gauge >= 0.2.5 && < 0.3,
-    deepseq >= 1.4,
+    deepseq >= 1.4.3,
```
sjakobi, line 200: Let's remove the now-redundant constraint

```suggestion
    deepseq,
```

sjakobi (commented):
I believe `--installed=-deepseq` needs to be added to the command that generates our Travis config, so it can use a non-bundled version.

https://github.com/haskell-unordered-containers/unordered-containers/blob/master/.travis.yml#L3

sjakobi:
I believe `bytestring` is the last package where the `installed` constraint needs to be removed. You can check this locally by using the generated `cabal.project.local` locally and building with GHC-8.0.2, GHC-7.10.3 and GHC-7.8.4.

sjakobi:
I'd like to cut a release fairly soon and ideally include this PR. Do you think you can finish this PR in the next, say, 2 weeks, @LightAndLight? Otherwise I can take a stab at it, but I'd probably take the CPP route.

LightAndLight:
@sjakobi I copied the `cabal.project.local` from the failing CI build and `cabal new-build -w {ghc-7.8.4,ghc-7.10.3,ghc-8.0.2}` all succeeded on my machine. I don't understand what's wrong with the CI build- can you help me?

sjakobi:
I can reproduce the error from https://travis-ci.com/github/haskell-unordered-containers/unordered-containers/jobs/495652251 by using the following `cabal.project.local`:

```
constraints: Cabal installed
constraints: array installed
constraints: base installed
constraints: directory installed
constraints: filepath installed
constraints: ghc installed
constraints: ghc-boot installed
constraints: ghc-boot-th installed
constraints: ghc-prim installed
constraints: ghci installed
constraints: haskeline installed
constraints: hoopl installed
constraints: hpc installed
constraints: integer-gmp installed
constraints: pretty installed
constraints: rts installed
constraints: template-haskell installed
constraints: terminfo installed
constraints: transformers installed
constraints: xhtml installed
```

and the following command:

```
cabal build -w ghc-8.0.2 --enable-benchmarks --enable-tests
```

I think the error message is pretty unhelpful, but I noticed that the `process` dependency might be constrained by its `directory` and `filepath` dependencies. Indeed removing these constraints from the `cabal.project.local` makes the command succeed.

With 7.10.3 I also have to remove the constraint on `pretty` and increase `--max-backjumps`.

I don't have 7.8.4 installed, so I haven't been able to test with it.

I'm not sure how much more pain it would take to fix the CI setup. We can probably take a shortcut and disable the benchmarks for GHC < 8.

I'm slightly worried that some users might end up going down this constraint-fighting path too though, so I think we should rather switch to the CPP route. I'm sorry that I didn't insist on it in the first place.

sjakobi:
Merged via #314. Thank you, @LightAndLight!

PR #406, opened 2022-04-09
Make intersections much faster

### LINKED ISSUE #225: Make intersections much faster

oberblastmeister:
Resolves #225. I also incorporated techniques from #291 and #362. All tests pass. I need to do some benchmarking and clean up the code, probably reformat it also.

### DIFF
--- a/Data/HashMap/Internal.hs
+++ b/Data/HashMap/Internal.hs
@@ -819,17 +819,9 @@ insertNewKey !h0 !k0 x0 !m0 = go h0 k0 x0 0 m0
         in Full (update32 ary i st')
       where i = index h s
     go h k x s t@(Collision hy v)
-        | h == hy   = Collision h (snocNewLeaf (L k x) v)
+        | h == hy   = Collision h (A.snoc v (L k x))
         | otherwise =
             go h k x s $ BitmapIndexed (mask hy s) (A.singleton t)
-      where
-        snocNewLeaf :: Leaf k v -> A.Array (Leaf k v) -> A.Array (Leaf k v)
-        snocNewLeaf leaf ary = A.run $ do
-          let n = A.length ary
-          mary <- A.new_ (n + 1)
-          A.copy ary 0 mary 0 n
-          A.write mary n leaf
-          return mary
 {-# NOINLINE insertNewKey #-}
 
 
@@ -1008,12 +1000,8 @@ insertModifyingArr :: Eq k => v -> (v -> (# v #)) -> k -> A.Array (Leaf k v)
 insertModifyingArr x f k0 ary0 = go k0 ary0 0 (A.length ary0)
   where
     go !k !ary !i !n
-        | i >= n = A.run $ do
-            -- Not found, append to the end.
-            mary <- A.new_ (n + 1)
-            A.copy ary 0 mary 0 n
-            A.write mary n (L k x)
-            return mary
+          -- Not found, append to the end.
+        | i >= n = A.snoc ary $ L k x
         | otherwise = case A.index ary i of
             (L kx y) | k == kx   -> case f y of
                                       (# y' #) -> if ptrEq y y'
@@ -1780,21 +1768,25 @@ intersectionWith f = inline intersectionWithKey $ const f
 -- | /O(n*log m)/ Intersection of two maps. If a key occurs in both maps
 -- the provided function is used to combine the values from the two
 -- maps.
-intersectionWithKey ::  (Eq k, Hashable k) => (k -> v1 -> v2 -> v3) -> HashMap k v1 -> HashMap k v2 -> HashMap k v3
-intersectionWithKey f = go 0
+intersectionWithKey :: (Eq k, Hashable k) => (k -> v1 -> v2 -> v3) -> HashMap k v1 -> HashMap k v2 -> HashMap k v3
+intersectionWithKey f = intersectionWithKey# (\k v1 v2 -> (# f k v1 v2 #))
+{-# INLINABLE intersectionWithKey #-}
+
+intersectionWithKey# :: Eq k => (k -> v1 -> v2 -> (# v3 #)) -> HashMap k v1 -> HashMap k v2 -> HashMap k v3
+intersectionWithKey# f = go 0
   where
     -- empty vs. anything
     go !_ _ Empty = Empty
     go _ Empty _ = Empty
     -- leaf vs. anything
-    go s (Leaf h1 (L k1 v1)) t2 = lookupCont (\_ -> Empty) (\v _ -> Leaf h1 $ L k1 $ f k1 v1 v) h1 k1 s t2
-    go s t1 (Leaf h2 (L k2 v2)) = lookupCont (\_ -> Empty) (\v _ -> Leaf h2 $ L k2 $ f k2 v v2) h2 k2 s t1
+    go s (Leaf h1 (L k1 v1)) t2 = lookupCont (\_ -> Empty) (\v _ -> case f k1 v1 v of (# v' #) -> Leaf h1 $ L k1 v') h1 k1 s t2
+    go s t1 (Leaf h2 (L k2 v2)) = lookupCont (\_ -> Empty) (\v _ -> case f k2 v v2 of (# v' #) -> Leaf h2 $ L k2 v') h2 k2 s t1
     -- collision vs. collision
     go _ (Collision h1 ls1) (Collision h2 ls2)
       | h1 == h2 = if A.length ls == 0 then Empty else Collision h1 ls
       | otherwise = Empty
       where
-        ls = intersectionUnorderedArrayWithKey (\k v1 v2 -> (# f k v1 v2 #)) ls1 ls2
+        ls = intersectionUnorderedArrayWithKey f ls1 ls2
     -- branch vs. branch
     go s (BitmapIndexed b1 ary1) (BitmapIndexed b2 ary2) = intersectionArray s b1 b2 ary1 ary2
     go s (BitmapIndexed b1 ary1) (Full ary2) = intersectionArray s b1 fullNodeMask ary1 ary2
@@ -1830,7 +1822,7 @@ intersectionWithKey f = go 0
     normalize b ary
       | A.length ary == 0 = Empty
       | otherwise = bitmapIndexedOrFull b ary
-{-# INLINABLE intersectionWithKey #-}
+{-# INLINE intersectionWithKey# #-}
 
 intersectionArrayBy :: (v1 -> v2 -> HashMap k v) -> Bitmap -> Bitmap -> A.Array v1 -> A.Array v2 -> (Bitmap, A.Array (HashMap k v))
 intersectionArrayBy f = intersectionArrayByFilter f $ \case Empty -> False; _ -> True
@@ -2263,12 +2255,8 @@ updateOrSnocWithKey :: Eq k => (k -> v -> v -> (# v #)) -> k -> v -> A.Array (Le
 updateOrSnocWithKey f k0 v0 ary0 = go k0 v0 ary0 0 (A.length ary0)
   where
     go !k v !ary !i !n
-        | i >= n = A.run $ do
-            -- Not found, append to the end.
-            mary <- A.new_ (n + 1)
-            A.copy ary 0 mary 0 n
-            A.write mary n (L k v)
-            return mary
+        -- Not found, append to the end.
+        | i >= n = A.snoc ary $ L k v
         | L kx y <- A.index ary i
         , k == kx
         , (# v2 #) <- f k v y
--- a/Data/HashMap/Internal.hs
+++ b/Data/HashMap/Internal.hs
@@ -1784,8 +1784,16 @@ intersectionWithKey# f = go 0
     go !_ _ Empty = Empty
     go _ Empty _ = Empty
     -- leaf vs. anything
-    go s (Leaf h1 (L k1 v1)) t2 = lookupCont (\_ -> Empty) (\v _ -> case f k1 v1 v of (# v' #) -> Leaf h1 $ L k1 v') h1 k1 s t2
-    go s t1 (Leaf h2 (L k2 v2)) = lookupCont (\_ -> Empty) (\v _ -> case f k2 v v2 of (# v' #) -> Leaf h2 $ L k2 v') h2 k2 s t1
+    go s (Leaf h1 (L k1 v1)) t2 =
+      lookupCont
+        (\_ -> Empty)
+        (\v _ -> case f k1 v1 v of (# v' #) -> Leaf h1 $ L k1 v')
+        h1 k1 s t2
+    go s t1 (Leaf h2 (L k2 v2)) =
+      lookupCont
+        (\_ -> Empty)
+        (\v _ -> case f k2 v v2 of (# v' #) -> Leaf h2 $ L k2 v')
+        h2 k2 s t1
     -- collision vs. collision
     go _ (Collision h1 ls1) (Collision h2 ls2) = intersectionCollisions f h1 h2 ls1 ls2
     -- branch vs. branch
@@ -1827,29 +1835,29 @@ intersectionArrayBy ::
 intersectionArrayBy f !b1 !b2 !ary1 !ary2
   | b1 .&. b2 == 0 = Empty
   | otherwise = runST $ do
-  mary <- A.new_ $ popCount bIntersect
-  -- iterate over nonzero bits of b1 .|. b2
-  let go !i !i1 !i2 !b !bFinal
-        | b == 0 = pure (i, bFinal)
-        | testBit $ b1 .&. b2 = do
-          x1 <- A.indexM ary1 i1
-          x2 <- A.indexM ary2 i2
-          case f x1 x2 of
-            Empty -> go i (i1 + 1) (i2 + 1) b' (bFinal .&. complement m)
-            _ -> do
-              A.write mary i $! f x1 x2
-              go (i + 1) (i1 + 1) (i2 + 1) b' bFinal
-        | testBit b1 = go i (i1 + 1) i2 b' bFinal
-        | otherwise = go i i1 (i2 + 1) b' bFinal
-        where
-          m = 1 `unsafeShiftL` countTrailingZeros b
-          testBit x = x .&. m /= 0
-          b' = b .&. complement m
-  (len, bFinal) <- go 0 0 0 bCombined bIntersect
-  case len of
-    0 -> pure Empty
-    1 -> A.read mary 0
-    _ -> bitmapIndexedOrFull bFinal <$> (A.unsafeFreeze =<< A.shrink mary len)
+    mary <- A.new_ $ popCount bIntersect
+    -- iterate over nonzero bits of b1 .|. b2
+    let go !i !i1 !i2 !b !bFinal
+          | b == 0 = pure (i, bFinal)
+          | testBit $ b1 .&. b2 = do
+            x1 <- A.indexM ary1 i1
+            x2 <- A.indexM ary2 i2
+            case f x1 x2 of
+              Empty -> go i (i1 + 1) (i2 + 1) b' (bFinal .&. complement m)
+              _ -> do
+                A.write mary i $! f x1 x2
+                go (i + 1) (i1 + 1) (i2 + 1) b' bFinal
+          | testBit b1 = go i (i1 + 1) i2 b' bFinal
+          | otherwise = go i i1 (i2 + 1) b' bFinal
+          where
+            m = 1 `unsafeShiftL` countTrailingZeros b
+            testBit x = x .&. m /= 0
+            b' = b .&. complement m
+    (len, bFinal) <- go 0 0 0 bCombined bIntersect
+    case len of
+      0 -> pure Empty
+      1 -> A.read mary 0
+      _ -> bitmapIndexedOrFull bFinal <$> (A.unsafeFreeze =<< A.shrink mary len)
   where
     bCombined = b1 .|. b2
     bIntersect = b1 .&. b2
@@ -1857,25 +1865,25 @@ intersectionArrayBy f !b1 !b2 !ary1 !ary2
 
 intersectionCollisions :: Eq k => (k -> v1 -> v2 -> (# v3 #)) -> Hash -> Hash -> A.Array (Leaf k v1) -> A.Array (Leaf k v2) -> HashMap k v3
 intersectionCollisions f h1 h2 ary1 ary2
-  | h1 == h2 =  runST $ do
-  mary2 <- A.thaw ary2 0 $ A.length ary2
-  mary <- A.new_ $ min (A.length ary1) (A.length ary2)
-  let go i j
-        | i >= A.length ary1 || j >= A.lengthM mary2 = pure j
-        | otherwise = do
-          L k1 v1 <- A.indexM ary1 i
-          searchSwap k1 j mary2 >>= \case
-            Just (L _k2 v2) -> do
-              let !(# v3 #) = f k1 v1 v2
-              A.write mary j $ L k1 v3
-              go (i + 1) (j + 1)
-            Nothing -> do
-              go (i + 1) j
-  len <- go 0 0
-  case len of
-    0 -> pure Empty
-    1 -> Leaf h1 <$> A.read mary 0
-    _ -> Collision h1 <$> (A.unsafeFreeze =<< A.shrink mary len)
+  | h1 == h2 = runST $ do
+    mary2 <- A.thaw ary2 0 $ A.length ary2
+    mary <- A.new_ $ min (A.length ary1) (A.length ary2)
+    let go i j
+          | i >= A.length ary1 || j >= A.lengthM mary2 = pure j
+          | otherwise = do
+            L k1 v1 <- A.indexM ary1 i
+            searchSwap k1 j mary2 >>= \case
+              Just (L _k2 v2) -> do
+                let !(# v3 #) = f k1 v1 v2
+                A.write mary j $ L k1 v3
+                go (i + 1) (j + 1)
+              Nothing -> do
+                go (i + 1) j
+    len <- go 0 0
+    case len of
+      0 -> pure Empty
+      1 -> Leaf h1 <$> A.read mary 0
+      _ -> Collision h1 <$> (A.unsafeFreeze =<< A.shrink mary len)
   | otherwise = Empty
 {-# INLINE intersectionCollisions #-}
 
--- a/Data/HashMap/Internal.hs
+++ b/Data/HashMap/Internal.hs
@@ -144,8 +144,8 @@ import Control.Applicative        (Const (..))
 import Control.DeepSeq            (NFData (..), NFData1 (..), NFData2 (..))
 import Control.Monad.ST           (ST, runST)
 import Data.Bifoldable            (Bifoldable (..))
-import Data.Bits                  (complement, countTrailingZeros, popCount,
-                                   unsafeShiftL, unsafeShiftR, (.&.), (.|.))
+import Data.Bits                  (complement, popCount, unsafeShiftL,
+                                   unsafeShiftR, (.&.), (.|.), countTrailingZeros, shiftL)
 import Data.Coerce                (coerce)
 import Data.Data                  (Constr, Data (..), DataType)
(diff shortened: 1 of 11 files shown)


oberblastmeister:
Benchmarks show that this is 2x faster than the previous implementation!

Data/HashMap/Internal.hs
```diff
@@ -163,9 +163,9 @@ import Text.Read                  hiding (step)
 import qualified Data.Data                   as Data
 import qualified Data.Foldable               as Foldable
 import qualified Data.Functor.Classes        as FC
-import qualified Data.HashMap.Internal.Array as A
 import qualified Data.Hashable               as H
 import qualified Data.Hashable.Lifted        as H
+import qualified Data.HashMap.Internal.Array as A
 import qualified Data.List                   as List
```
treeowl, line 169: Why are you rearranging the imports?

Data/HashMap/Internal.hs
```diff
@@ -210,6 +210,7 @@ data HashMap k v
     | Leaf !Hash !(Leaf k v)
     | Full !(A.Array (HashMap k v))
     | Collision !Hash !(A.Array (Leaf k v))
+    deriving (Show)
```
treeowl: This obviously needs to be reverted.

Data/HashMap/Internal.hs
```diff
@@ -1771,37 +1768,135 @@ differenceWith f a b = foldlWithKey' go empty a
 -- | /O(n*log m)/ Intersection of two maps. Return elements of the first
 -- map for keys existing in the second.
 intersection :: (Eq k, Hashable k) => HashMap k v -> HashMap k w -> HashMap k v
-intersection a b = foldlWithKey' go empty a
-  where
-    go m k v = case lookup k b of
-                 Just _ -> unsafeInsert k v m
-                 _      -> m
+intersection = intersectionWith const
 {-# INLINABLE intersection #-}
```
treeowl, line 1771: I think we probably want to explicitly `inline intersectionWithKey` here. We really don't want stupid thunks in the result. Another possibility might be to base them all (lazy and strict) on a version that takes `k -> v -> v -> (# v #)`, though that could be a bit slower.

Data/HashMap/Internal.hs
oberblastmeister: It looks like `stylish-haskell` does this?

treeowl (changes requested):
Just a few concerns, most importantly about inlining.

Data/HashMap/Internal.hs
treeowl: Please anti-do it. It's distracting.

oberblastmeister: I can anti-do it. Is there some configuration option for `stylish-haskell` though, I can't seem to find it.

Data/HashMap/Internal.hs
oberblastmeister: I would favor using unboxed tuples. What are some reasons why it would be slower?

sjakobi:
Thanks for working on this! I'll review on monday.

oberblastmeister ***@***.***> schrieb am Sa., 9. Apr. 2022,
17:10:

treeowl:
Is stylish Haskell happening on GitHub or something? If so, I guess @sjakobi wants that. Don't ask me; I don't know tools.

The potential performance concern with an unboxed tuple result is that when the passed function doesn't inline, we ended up with an extra function call per element. I don't know if that makes a big enough difference to worry about; it'd be worth experimenting. I imagine it would save a lot of code duplication if it turns out okay. Specifically, it would avoid source duplication between lazy and strict versions, and object code duplication among the variants.

oberblastmeister:
How is that different than with a function that does not return an unboxed tuple? Shouldn't it still have to make an extra function call for each element if the function doesn't inline?

treeowl:
Oh, I was thinking of not inlining the version that produces an unboxed tuple. If we *do* inline that into the other variants, then yeah, everything should be basically the same but with less source code. But it's worth finding out how much we pay for *not* inlining it, because object code isn't free either.

oberblastmeister:
What if we just inline the ones with the unboxed tuples. Is there any disadvantage? Because we already `inline intersectionWithKey`, so `inline intersectionWithKey#` shouldn't create extra code? Then later we can experiment and see if we don't have to inline the unboxed version.

treeowl:
I opened a PR against your branch to do that.

treeowl:
There are CI failures on older GHC, because shrinking wasn't available yet. We'll need a fall-back.

treeowl:
The fallback should probably just define the shrinking operation manually in the `Array` module. It won't be too efficient, but whatever.

oberblastmeister:
Should I just use something like `copy` to implement the fallback?

treeowl:
I dunno. I haven't actually looked at how you're using `shrink`. Really, you can do whatever you think is reasonable, but I'd prefer to keep the CPP for it in `Array` if possible.

treeowl:
Oh, I just looked. You'll want to use `cloneSmallMutableArray#`

oberblastmeister:
Something like
```haskell
#if __GLASGOW_HASKELL__ >= 8.10.7
shrink = ...
#else
shrink = ...
#endif
```
(never used cpp before)

treeowl:
Yup! It's one of the world's worst macro systems, but it's a lot faster than Template Haskell. *Sigh*

Data/HashMap/Internal/Array.hs
```diff
@@ -204,6 +209,23 @@ new _n@(I# n#) b =
...
+  CHECK_GE("shrink", _n, (0 :: Int))
+  CHECK_LE("shrink", _n, (lengthM mary))
+  ST $ \s -> case Exts.shrinkSmallMutableArray# (unMArray mary) n# s of
+    s' -> (# s', mary #)
+#else
+shrink mary n = do
+  mary' <- new_ n
+  copyM mary 0 mary' 0 n
+  pure mary'
+#endif
```
treeowl, line 226: I really think you should use `cloneSmallMutableArray#` here to avoid filling the new array with errors first.

treeowl:
You probably have an off by one error in your clone; GHC has some funny ways of counting copies.

treeowl:
Non-urgent, non-merge-blocking: we might want to compare the performance of cloning to shrinking when using the non-moving garbage collector. Shrinking is rather more involved in that context, and it looks like it even makes a C call. I don't know if the effect will be magnified by generational issues for large maps, but it wouldn't shock me.

Data/HashMap/Internal.hs
```diff
@@ -1771,37 +1755,143 @@ differenceWith f a b = foldlWithKey' go empty a
...
+    -- leaf vs. anything
+    go s (Leaf h1 (L k1 v1)) t2 = lookupCont (\_ -> Empty) (\v _ -> case f k1 v1 v of (# v' #) -> Leaf h1 $ L k1 v') h1 k1 s t2
+    go s t1 (Leaf h2 (L k2 v2)) = lookupCont (\_ -> Empty) (\v _ -> case f k2 v v2 of (# v' #) -> Leaf h2 $ L k2 v') h2 k2 s t1
+    -- collision vs. collision
+    go _ (Collision h1 ls1) (Collision h2 ls2)
+      | h1 == h2 = runST $ do
+        (len, mary) <- intersectionUnorderedArrayWithKey f ls1 ls2
+        if len == 0
+          then pure Empty
+          else Collision h1 <$> (A.unsafeFreeze =<< A.shrink mary len)
```
treeowl, line 1790: My comment seems to have gone into the void. We need to deal with `len=1` (`Leaf`). I'm also concerned about the tests not seeming to catch the mistake.

oberblastmeister:
Unfortunately, adding `{-# NOINLINE -#}` to `intersectionWithKey#` results in a 10 microsecond performance degredation. So it looks like we will have to inline it to eliminate the closure

treeowl:
We don't want to `NOINLINE` it; that loses specialization. The question is whether it's okay for it to be `INLINABLE` and not explicitly `inline`d. We'd want to verify that it doesn't get inlined into the benchmark.

treeowl:
It may well need inlining; just want to double check. And that percentage change is it?

oberblastmeister:
Putting `INLINABLE` results in the same performance decrease

oberblastmeister:
Sorry, `INLINEABLE` is slower than `INLINE`, but faster than `NOINLINE`, probably due to the typeclass specialization that you described

oberblastmeister:
|`INLINE`|`INLINABLE`|`NOINLINE`
|---|---|---|
|40|50|56|

Data/HashMap/Internal.hs
oberblastmeister: I fixed this, but I am not sure what to do about the tests

treeowl:
Okay, I guess we should go with `INLINE` for now. I guess in practice it's not likely to be called with many different function arguments in a given program. We should make sure that GHC sees it as having arity 2—it should inline once applied to a dictionary and a function (and similarly for the user-exposed functions). Do you want to do the strict thing now, or in another PR?

oberblastmeister:
It is pretty late for me now, so I could either wait until tomorrow or do it in another pr.

treeowl:
Any thoughts, @sjakobi ?

sjakobi:
I'm AFK. I'll Review tomorrow.

David Feuer ***@***.***> schrieb am So., 10. Apr. 2022, 03:31:

oberblastmeister:
Why is `unionWith`, `unionWithKey`, and `union` marked as `INLINE`? Shouldn't they be marked as `INLINABLE` and then use the `inline` function explicitly when we want to implement something in terms of the other.  For example `union = inline unionWith const` just like I am doing with `intersection`. Wouldn't marking them as `INLINE` create excessive code bloat?

sjakobi:
> Why is `unionWith`, `unionWithKey`, and `union` marked as `INLINE`?

`union` is actually marked `INLINABLE`.

For `unionWith[Key]`, I suspect the motivation is to inline the function argument, so using it doesn't incur function calls.

We can consider using `INLINABLE` for these instead. Feel free to open an issue.

Data/HashMap/Internal.hs
```diff
@@ -1771,37 +1756,147 @@ differenceWith f a b = foldlWithKey' go empty a
...
+          testBit x = x .&. m /= 0
+          b' = b .&. complement m
+  (maryLen, bFinal) <- go 0 0 0 bCombined bIntersect
+  pure (bFinal, maryLen, mary)
+  where
+    bCombined = b1 .|. b2
+    bIntersect = b1 .&. b2
+{-# INLINE intersectionArrayBy #-}
+
+intersectionUnorderedArrayWithKey :: Eq k => (k -> v1 -> v2 -> (# v3 #)) -> A.Array (Leaf k v1) -> A.Array (Leaf k v2) -> ST s (Int, A.MArray s (Leaf k v3))
```
sjakobi, line 1866: How about something shorter like this:

```suggestion
intersectionCollisions :: Eq k => (k -> v1 -> v2 -> (# v3 #)) -> A.Array (Leaf k v1) -> A.Array (Leaf k v2) -> ST s (Int, A.MArray s (Leaf k v3))
```

Data/HashMap/Internal.hs
```diff
@@ -1771,37 +1756,147 @@ differenceWith f a b = foldlWithKey' go empty a
...
+              let !(# v3 #) = f k1 v1 v2
+              A.write mary j $ L k1 v3
+              go (i + 1) (j + 1)
+            Nothing -> do
+              go (i + 1) j
+  maryLen <- go 0 0
+  pure (maryLen, mary)
+{-# INLINE intersectionUnorderedArrayWithKey #-}
+
+searchSwap :: Eq k => k -> Int -> A.MArray s (Leaf k v) -> ST s (Maybe (Leaf k v))
```
sjakobi, line 1885: Please add haddocks.

sjakobi (commented):
First batch of comments. I'll review more tomorrow.

treeowl:
When we *don't* inline the function, it's probably better to pass the function around through the recursion instead of pulling it out. Tradeoffs....

sjakobi:
Could you please rebase, so the changes from #407 are removed from the diff?

Data/HashMap/Internal.hs
```diff
@@ -1771,37 +1756,147 @@ differenceWith f a b = foldlWithKey' go empty a
...
+            _ -> do
+              A.write mary i $! f x1 x2
+              go (i + 1) (i1 + 1) (i2 + 1) b' bFinal
+        | testBit b1 = go i (i1 + 1) i2 b' bFinal
+        | otherwise = go i i1 (i2 + 1) b' bFinal
+        where
+          m = 1 `unsafeShiftL` countTrailingZeros b
+          testBit x = x .&. m /= 0
+          b' = b .&. complement m
+  (maryLen, bFinal) <- go 0 0 0 bCombined bIntersect
```
sjakobi, line 1859: The comment at the top seems incorrect: Currently the loop actually iterates over `b1 .|. b2`. It would be nice to change this though. In that case the `i1` and `i2` indices could be computed with `sparseIndex`.

sjakobi (commented):
I haven't reviewed `intersectionCollisions` yet. Could you add some documentation on `searchSwap` first?

Data/HashMap/Internal/Array.hs
```diff
@@ -204,6 +207,20 @@ new _n@(I# n#) b =
 new_ :: Int -> ST s (MArray s a)
 new_ n = new n undefinedElem
 
+-- when shrinkSmallMutableArray# is available, the returned array is the same as the array given, as it is shrunk in place
+-- otherwise a copy is made
+shrink :: MArray s a -> Int -> ST s (MArray s a)
+#if MIN_VERSION_GLASGOW_HASKELL(8, 10, 7, 0)
```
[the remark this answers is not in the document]
sjakobi: In fact the current pragma may not work correctly in some GHC releases due to https://gitlab.haskell.org/ghc/ghc/-/issues/21076.

Data/HashMap/Internal.hs
```diff
@@ -1771,37 +1756,156 @@ differenceWith f a b = foldlWithKey' go empty a
...
+  pure (bFinal, maryLen, mary)
+  where
+    bCombined = b1 .|. b2
+    bIntersect = b1 .&. b2
+{-# INLINE intersectionArrayBy #-}
+
+intersectionCollisions :: Eq k => (k -> v1 -> v2 -> (# v3 #)) -> A.Array (Leaf k v1) -> A.Array (Leaf k v2) -> ST s (Int, A.MArray s (Leaf k v3))
+intersectionCollisions f ary1 ary2 = do
+  mary2 <- A.thaw ary2 0 $ A.length ary2
```
sjakobi, line 1868: I wonder whether we actually need to allocate two arrays for this. The alternative would be to perform the search-and-swap operations on the output array itself.

It might be a bit tricky though – maybe leave it for a follow-up PR, so this one doesn't get too huge.

oberblastmeister: I think the issue with this is that the type could change. For example if we have two arrays with the numbers as keys, and the arrays are both different types
`1 2 3 4`
`3 4 2 1`
Let's thaw the first array, and mutate it to
`(f 3 3) 2 1 4`
`f 3 3` could change the type to be something difference than the `2 1 4`.

sjakobi: Ah, yes, good point. Unsafe coercions _might_ work for this, but I'd prefer not trying this in this PR.

treeowl: This is only an issue for `intersectionWithKey` and such; `intersection` itself has no type issue.

oberblastmeister:
What should I do about inlining? I understand the need to eliminate the closures, but the functions are truly massive, `intersection` has 1,700 terms, while `unionWithKey` has 2,200! Wouldn't it be bad to mark these as `{-# INLINE #-}`? We could add a comment saying to explicitly `inline` if you want to remove the closure.

Data/HashMap/Internal.hs
treeowl: There's another thing about `intersection` that's special: we can reuse the leaves.

Data/HashMap/Internal.hs
```diff
@@ -1771,37 +1756,156 @@ differenceWith f a b = foldlWithKey' go empty a
...
+  mary <- A.new_ $ A.length ary1 + A.length ary2
```
treeowl: I'm confused. Shouldn't the result size be bounded by `min (A.length ary1) (A.length ary2)`?

oberblastmeister:
Also if we pass the function around through recursion, then we wouldn't be able to implement `intersection` in terms of `intersectionWithKey` right?

Data/HashMap/Internal.hs
```diff
@@ -1771,37 +1756,156 @@ differenceWith f a b = foldlWithKey' go empty a
...
+  let go i j
+        | i >= A.length ary1 || j >= A.lengthM mary2 = pure j
```
treeowl, line 1875: I imagine it's better to use the known length of `mary2` than to calculate it here, to avoid potentially dragging in an extra cache line.

Data/HashMap/Internal.hs
```diff
@@ -1771,37 +1756,156 @@ differenceWith f a b = foldlWithKey' go empty a
...
+          L k1 v1 <- A.indexM ary1 i
+          searchSwap k1 j mary2 >>= \case
+            Just (L _k2 v2) -> do
+              let !(# v3 #) = f k1 v1 v2
+              A.write mary j $ L k1 v3
+              go (i + 1) (j + 1)
+            Nothing -> do
+              go (i + 1) j
+  maryLen <- go 0 0
+  pure (maryLen, mary)
```
treeowl, line 1886: Should we go ahead and produce the actual `HashMap` fragment here?

treeowl: > Also if we pass the function around through recursion, then we wouldn't be able to implement `intersection` in terms of `intersectionWithKey` right?

Right. There are dizzyingly many options with equally many trade-offs. I hate it. That said, I wonder if we should be implementing `intersection` separately anyway, since it can get sharing in the leaves. Ugh.

Data/HashMap/Internal.hs
sjakobi: Yes, it would be better if `intersection` had custom code for handling collisions. Maybe this can be achieved by changing `intersectionWithKey#` to something similar to `filterMapAux`.

I'd slightly prefer if we'd leave this for a follow-up PR though.

sjakobi:
> What should I do about inlining? I understand the need to eliminate the closures, but the functions are truly massive, `intersection` has 1,700 terms, while `unionWithKey` has 2,200! Wouldn't it be bad to mark these as `{-# INLINE #-}`?

I think we should stick with `INLINABLE` until we're convinced that `INLINE` is better somehow. `INLINABLE` is better for compile times for example.

> We could add a comment saying to explicitly `inline` if you want to remove the closure.

I think the docs of these functions are the wrong place to teach people about `GHC.Exts.inline`. Maybe it should be mentioned in https://github.com/input-output-hk/hs-opt-handbook.github.io?!

Data/HashMap/Internal.hs
sjakobi: The current code is already using `min`.

Data/HashMap/Internal.hs
sjakobi: > Should we go ahead and produce the actual `HashMap` fragment here?

Yeah, I think that would be clearer.

Data/HashMap/Internal.hs
sjakobi: We have https://github.com/haskell-unordered-containers/unordered-containers/issues/366 for tracking the tests improvements.

Data/HashMap/Internal.hs
treeowl: Oh, no, it won't bring an extra cache line; sorry.

treeowl: Anyway, this is in collision buckets, so it doesn't matter. Silly me.

Data/HashMap/Internal.hs
sjakobi: > In that case the i1 and i2 indices could be computed with sparseIndex.

For comparison, here is a version of  `unionArrayBy` that uses `sparseIndex` to compute all the indices:

https://github.com/haskell-unordered-containers/unordered-containers/blob/a780a8d942716a9c6adccf0501c0f7a949f64211/Data/HashMap/Internal.hs#L1626-L1647

I expect that keeping `i` as a loop argument will be more efficient than recomputing it on each iteration though.

Data/HashMap/Internal.hs
```diff
@@ -1771,37 +1760,149 @@ differenceWith f a b = foldlWithKey' go empty a
...
+    -- leaf vs. anything
+    go s (Leaf h1 (L k1 v1)) t2 = lookupCont (\_ -> Empty) (\v _ -> case f k1 v1 v of (# v' #) -> Leaf h1 $ L k1 v') h1 k1 s t2
+    go s t1 (Leaf h2 (L k2 v2)) = lookupCont (\_ -> Empty) (\v _ -> case f k2 v v2 of (# v' #) -> Leaf h2 $ L k2 v') h2 k2 s t1
+    -- collision vs. collision
+    go _ (Collision h1 ls1) (Collision h2 ls2) = intersectionCollisions f h1 h2 ls1 ls2
+    -- branch vs. branch
+    go s (BitmapIndexed b1 ary1) (BitmapIndexed b2 ary2) = intersectionArrayBy (go (s + bitsPerSubkey)) b1 b2 ary1 ary2
+    go s (BitmapIndexed b1 ary1) (Full ary2) = intersectionArrayBy (go (s + bitsPerSubkey)) b1 fullNodeMask ary1 ary2
+    go s (Full ary1) (BitmapIndexed b2 ary2) = intersectionArrayBy (go (s + bitsPerSubkey)) fullNodeMask b2 ary1 ary2
+    go s (Full ary1) (Full ary2) = intersectionArrayBy (go (s + bitsPerSubkey)) fullNodeMask fullNodeMask ary1 ary2
```
sjakobi, line 1795: Can you break these long lines here?

Data/HashMap/Internal.hs
```diff
@@ -1771,37 +1760,149 @@ differenceWith f a b = foldlWithKey' go empty a
...
+          testBit x = x .&. m /= 0
+          b' = b .&. complement m
+  (len, bFinal) <- go 0 0 0 bCombined bIntersect
+  case len of
+    0 -> pure Empty
+    1 -> A.read mary 0
+    _ -> bitmapIndexedOrFull bFinal <$> (A.unsafeFreeze =<< A.shrink mary len)
+  where
+    bCombined = b1 .|. b2
+    bIntersect = b1 .&. b2
```
sjakobi, line 1855: This `do`-block should be indented relatively to the `otherwise` guard.

treeowl: What do you mean?

sjakobi: I think it's just confusing that the `do`-block is not indented.

Instead of

```
  | otherwise = runST $ do
  mary <- A.new_ $ popCount bIntersect
  ...
```

It should

be

```
  | otherwise = runST $ do
      mary <- A.new_ $ popCount bIntersect
      ...
```

treeowl: Yeah, confusing enough that I didn't *see* that `do` block.

Data/HashMap/Internal.hs
oberblastmeister: Using `sparseIndex` makes benchmarks slower

sjakobi: Can you show me the diff? 

> Using sparseIndex makes benchmarks slower

By how much? I also think the benchmark data might be a bit weird.

oberblastmeister: How do I show you the diff?

sjakobi: Maybe just push it, and push a revert if you think that's better.

Data/HashMap/Internal.hs
oberblastmeister: Here is the branch https://github.com/oberblastmeister/unordered-containers/tree/fast-intersection-sparseIndex

Data/HashMap/Internal.hs
sjakobi: > It's slower by 10 microseconds

Can you paste the whole output for this benchmark?

oberblastmeister: You mean the one with `sparseIndex`?

sjakobi: Ideally both.

oberblastmeister: Sparse index:
```
All
  HashMap
    intersection
      Int:        OK (0.34s)
        52.1 μs ± 3.3 μs
      ByteString: OK (0.26s)
        62.6 μs ± 6.1 μs
```

oberblastmeister: Without sparse index:
```
All
  HashMap
    intersection
      Int:        OK (0.86s)
        42.1 μs ± 1.5 μs
      ByteString: OK (0.33s)
        47.7 μs ± 2.8 μs
```

sjakobi: Alright, thanks! I think we might get different results with data where there's less overlap between the two maps. But that can be investigated at a different time.

oberblastmeister:
@sjakobi Is there anything else that I need to do?

Data/HashMap/Internal.hs
```diff
@@ -143,16 +144,16 @@ import Control.Applicative        (Const (..))
...
+                                   unsafeShiftL, unsafeShiftR, (.&.), (.|.))
 import Data.Coerce                (coerce)
 import Data.Data                  (Constr, Data (..), DataType)
 import Data.Functor.Classes       (Eq1 (..), Eq2 (..), Ord1 (..), Ord2 (..),
                                    Read1 (..), Show1 (..), Show2 (..))
 import Data.Functor.Identity      (Identity (..))
-import Data.HashMap.Internal.List (isPermutationBy, unorderedCompare)
 import Data.Hashable              (Hashable)
 import Data.Hashable.Lifted       (Hashable1, Hashable2)
+import Data.HashMap.Internal.List (isPermutationBy, unorderedCompare)
```
sjakobi, line 156: FWIW, the changed sorting of imports is probably due to https://github.com/haskell/stylish-haskell/pull/385, which was recently released.

sjakobi (approved):
Can you fix the merge conflict, @oberblastmeister?

I'll merge afterwards.

Data/HashMap/Internal.hs
sjakobi: I have recorded these ideas in https://github.com/haskell-unordered-containers/unordered-containers/issues/415.

Data/HashMap/Internal.hs
sjakobi: Follow-up issue in https://github.com/haskell-unordered-containers/unordered-containers/issues/416.

sjakobi:
Thank you, @oberblastmeister! :)

PR #495, opened 2024-10-14
Remove up version limit of template-haskell and hashable

Shadasviar:
Limiting upper version of packages causes dependency hell. The only reason why `master` does not compile with `hashable 1.5.0.0` and `template-haskell 2.22.0.0`, that I use in my project, is upper limits for its versions. It would be better to limit upper version of libraries only when they change API in incompatible way.

If API of any dependency will change in incompatible way, `unordered-containers` will not compile anyway. If next versions of dependencies would not break API (that is most often scenario), `unordered-containers` will have compatibility for free, without updating upstream from `<22` to `<23` etc with every new dependency release.

### DIFF
--- a/unordered-containers.cabal
+++ b/unordered-containers.cabal
@@ -29,7 +29,8 @@ cabal-version:  >=1.10
 extra-source-files: CHANGES.md
 
 tested-with:
-  GHC ==9.8.1
+  GHC ==9.10.1
+   || ==9.8.1
    || ==9.6.3
    || ==9.4.7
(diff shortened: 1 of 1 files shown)


andreasabel:
Yes, I agree, personally.
I'd argue in exactly the same way.

But I guess the policy of the maintainers is to have upper bounds strictly compatible with the PVP.

andreasabel:
I think this PR should be closed now that the respective bounds have been bumped:
- #499

PR #544, opened 2025-11-09
Make it compile with MicroHs

augustss:
The title says it all.

I can add MicroHs CI if you want.

### DIFF
--- a/Data/HashMap/Internal.hs
+++ b/Data/HashMap/Internal.hs
@@ -709,7 +709,7 @@ lookupCont ::
   -> k
   -> Shift
   -> HashMap k v -> r
-lookupCont absent present !h0 !k0 !s0 !m0 = go h0 k0 s0 m0
+lookupCont absent present !h0 !k0 !s0 m0 = go h0 k0 s0 m0
   where
     go :: Eq k => Hash -> k -> Shift -> HashMap k v -> r
     go !_ !_ !_ Empty = absent (# #)
@@ -846,7 +846,7 @@ insert' h0 k0 v0 m0 = go h0 k0 v0 0 m0
 --  - the key equality check on a Leaf
 --  - check for its existence in the array for a hash collision
 insertNewKey :: Hash -> k -> v -> HashMap k v -> HashMap k v
-insertNewKey !h0 !k0 x0 !m0 = go h0 k0 x0 0 m0
+insertNewKey !h0 !k0 x0 m0 = go h0 k0 x0 0 m0
   where
     go !h !k x !_ Empty = Leaf h (L k x)
     go h k x s t@(Leaf hy l)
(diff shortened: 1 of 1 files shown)


Data/HashMap/Internal.hs
```diff
@@ -181,6 +181,10 @@ import qualified Data.List                   as List
 import qualified GHC.Exts                    as Exts
 import qualified Language.Haskell.TH.Syntax  as TH
 
+#if defined(__MHS__)
+import Data.Traversable
+#endif
```
treeowl: This is okay, but also blecherous. We actually want to switch on the "base library" variant, not the compiler. Do you know if there's been any progress on how to do that?

Data/HashMap/Internal.hs
```diff
@@ -914,10 +924,11 @@ setAtPosition i k x ary = A.update ary i (L k x)
 
 
 -- | In-place update version of insert
-unsafeInsert :: (Eq k, Hashable k) => k -> v -> HashMap k v -> HashMap k v
+unsafeInsert :: forall k v. (Eq k, Hashable k) => k -> v -> HashMap k v -> HashMap k v
```
treeowl: Why do we need this? Do you implement the "let should not be generalized" paper unconditionally?

Data/HashMap/Internal.hs
```diff
@@ -191,9 +195,11 @@ data Leaf k v = L !k v
 instance (NFData k, NFData v) => NFData (Leaf k v) where
     rnf (L k v) = rnf k `seq` rnf v
 
+#if !defined(__MHS__)
```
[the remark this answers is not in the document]
augustss: Sure, if you prefer.

Data/HashMap/Internal.hs
treeowl: I think the condition is totally acceptable for this PR; I just think it would be good to open up a community discussion about the broader issue.

Data/HashMap/Internal.hs
augustss: Yes, MonoLocalBinds is always on. 
It's usually not a problem, but for runST its annoying.

Data/HashMap/Internal.hs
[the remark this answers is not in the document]
treeowl: I think that's better, because TH is a GHC-specific feature. MHS isn't "weird" for not supporting it.

augustss:
MicroHs does not have unlifted types. But it supports the syntax, and you get the lifted types instead. Of course, that's not as efficient, but at least it allows more packages to build (e.g. async needs u-c)

There's a new package for MicroHs called ghc-compat which implements parts of ghc specific modules. I will flesh it out as I need more things.

treeowl:
Would you please rebase?

PR #558, opened 2025-11-26
Remove redundant `Eq` constraints

sjakobi:
`Eq` has been a superclass of `Hashable` since `hashable-1.4`, which is the minimum version u-c supports.


sjakobi:
> I don't like that the `Eq` dict is allocated and passed as a thunk though. Do you know how this can be avoided?

Maybe [`-fspec-eval-dictfun`](https://downloads.haskell.org/ghc/latest/docs/users_guide/using-optimisation.html#ghc-flag-fspec-eval-dictfun) would help. Will give it a try.

sjakobi:
So for non-specialized uses of u-c this is possibly still a mixed bag, but I doubt it's a large effect one way or the other.

PR #563, opened 2025-11-29
Relax constraints of `HashSet.difference` and `Strict.differenceWith`

sjakobi:
I forgot to do this in #535 and #542.


treeowl:
This needs a major version bump and change log entry as well.

sjakobi:
> This needs a major version bump and change log entry as well.

We've previously included similar changes in minor version bumps. See e.g. https://github.com/haskell-unordered-containers/unordered-containers/issues/443

What changed since 2022 that would require a major version bump now?

sjakobi:
Reverted via https://github.com/haskell-unordered-containers/unordered-containers/commit/d3d9a72d071dfced624f1011d19971404322edfe.

PR #570, opened 2025-12-05
WIP: Add `nub`

### LINKED ISSUE #560: `hashNub` (and `hashNubOn`)
I just came across this blogpost comparing different versions of `nub`: https://oleg.fi/gists/posts/2021-01-07-discrimination-benchmarks.html

(Correspondig benchmarks: https://github.com/ekmett/discrimination/pull/33/files)

Hoogle finds several implementations of `hashNub` in several packages: https://hoogle.haskell.org/?hoogle=hashNub

So maybe u-c should export the "definitive" version of that function, similar to how `containers` exports [`nubOrd[On]`](https://hackage-content.haskell.org/package/containers-0.8/docs/Data-Containers-ListUtils.html#v:nubOrd)?!

(Bikeshed: Would `nubHash[On]` be better than `hashNub[On]`?!)

I _think_ we can use `unsafeInsert` instead of `insert` (or at least `insertNewKey`), so this version should be even faster.

In `witherable`, there's a more general version of [`hashNub`](https://hackage.haskell.org/package/witherable-0.5/docs/Witherable.html#v:hashNub) and also a [`hashNubOn`](https://hackage.haskell.org/package/witherable-0.5/docs/Witherable.html#v:hashNubOn) based on `alterF`. Maybe we can help improve these versions too.

---

(All this makes me curious how a `hashNub` based on [`vector-hashtables`](https://hackage.haskell.org/package/vector-hashtables) would fare.)

CC: @phadej, @klapaucius

sjakobi:
Resolves #560.

### DIFF
--- a/Data/HashMap/Internal.hs
+++ b/Data/HashMap/Internal.hs
@@ -2933,9 +2933,10 @@ instance Hashable k => Exts.IsList (HashMap k v) where
     toList   = toList
 #endif
 
-nub :: Hashable a => [a] -> [a]
+nub :: forall a. Hashable a => [a] -> [a]
 nub = \l -> runST (nub_ l empty)
   where
+    nub_ :: forall s. [a] -> HashMap a () -> ST s [a]
     nub_ [] _seen = pure []
     nub_ (x:xs) seen
       | Just _ <- lookup' h x seen = nub_ xs seen


Data/HashMap/Internal.hs
```diff
@@ -994,6 +996,38 @@ unsafeInsert k0 v0 m0 = runST (go h0 k0 v0 0 m0)
         | otherwise = go h k x s $ BitmapIndexed (mask hy s) (A.singleton t)
 {-# INLINABLE unsafeInsert #-}
 
+unsafeInsertNewKeyM :: Hash -> k -> v -> HashMap k v -> ST s (HashMap k v)
```
treeowl: Why "new"?

treeowl (commented):
There are two situations with different needs:

1. Repetitions are common. Your current code is optimized for this case. 
2. Repetitions are rare.

In the rare-repetition case, we almost always insert the key, so checking first seems like a waste of time. A version optimized for this would return (something isomorphic to) `Maybe (HashMap k a)`. Since we don't have an efficient `setjmp/longjmp`, I believe this will be slower than your version in the common-repetition case.

Do we want two versions of the function?

PR #571, opened 2025-12-05
Refactor `delete`

sjakobi:


Data/HashMap/Internal.hs
```diff
@@ -1158,56 +1158,56 @@ unsafeInsertWithKey f k0 v0 m0 = runST (go h0 k0 v0 0 m0)
...
-deleteFromSubtree !_ !_ !_ Empty = Empty
-deleteFromSubtree h k _ t@(Leaf hy (L ky _))
-    | hy == h && ky == k = Empty
-    | otherwise          = t
-deleteFromSubtree h k s t@(BitmapIndexed b ary)
-    | b .&. m == 0 = t
-    | otherwise
-    = case A.index# ary i of
+deleteFromSubtree :: Eq k => Shift -> Hash -> k -> HashMap k v -> HashMap k v
+deleteFromSubtree !s !h !k = \case
```
sjakobi: I'm a bit unhappy about the forcing of keys. The upside is that key types likes `Int` and `ShortByteString` are unboxed. The downside is that key types that can't be unboxed to an unlifted representation are unnecessarily evaluated again and again.

Maybe we can use [`Strict#`](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/15054) for this, once it's released?!

treeowl: Where/why are keys getting re-forced?

sjakobi: Interesting! I agree with your reasoning but I doubt that GHC's strictness analyser could be convinced to believe it too! :smile:

treeowl: I bet it can, but I need to go back to sleep.

